PR #4, opened 2025-04-08
Initialize Frontend

MaxDouglas1991:
# Description

Set up the frontend for BC WAT
Implements Quasar and Mapbox


cetooley:
Please add a `.env.example` for constants required to run the client - please also include in `README.md` how to run the client - I assume `npm run dev` but it needs to be clearly outlined.

client/src/components/home/HomePage.vue
```diff
@@ -0,0 +1,14 @@
+<template>
+    <div class="home-container">
+        <h3>BC Water Availability Tool</h3>
+    </div>
+</template>
+
+<script setup></script>
+
+<style lang="scss" scoped>
+.home-container {
```
cetooley, line 10: is the home-container class used anywhere else? If not, the "scoped" prop is completely unecessary.

cetooley (commented):
Left some comments - we need to be proactive about comments in the source code, and we also need to be strict about using values that can affect global properties

cetooley (approved):
:shipit:

PR #6, opened 2025-04-14
Init Flask Api - Hello World - PyTest - Dockerized

LiamF-FS:
# Description

- Create Hello World RESTful API using Flask
- Dockerize
- Init Unit Tests

## Types of changes

## Checklist

## Further comments

Choosing to use a Flask API as it is simple to use, suits our use case, and has extensive helm support for deploying.


docker-compose.yaml
```diff
@@ -0,0 +1,7 @@
+version: '3.9'
+services:
+  backend:
+    build: './backend'
+    ports:
+      - "8000:8000"
```
AndersFarrFoundry, line 6: We should maybe add a TODO to add a docker-compose set up for the frontend?

AndersFarrFoundry (approved):
LGTM - thanks for setting this up.

docker-compose.yaml
LiamF-FS: Good shout, I can do this. Further discussions will need to be had regarding what else should go in the docker compose - IE, the airflow folder, and our db init is still up in the air on if we will have enough resources for clustered postgres.

PR #8, opened 2025-04-15
ETL pipeline Skeleton and implementation of first one

Kashike-Umemura:
# Description
Two tickets have been completed in this PR:
## BCWAT-67: Create a Skeleton of the ETL pipeline
This is the initial PR for the scrapers. The goal is to have a structure in place so that the implementation is in the style of "Fill in the Blank", albeit quite a bit more complicated than that.
In addition to the skeleton, an unit test skeleton has been implemented for the scrapers.

## BCWAT-68 Complete Scraper Pipeline 1
The first scraper to be implemented! Although not 100% complete, it is a good starting point. Further issues will be made for filling out the tests for this scrapers as well as the other functions that are in the child classes.

## Types of changes

## Checklist

- [ x] I have read the [CONTRIBUTING](CONTRIBUTING.md) doc
- [ x] I have added necessary documentation (if appropriate)

## Further comments

Did not check that the unit tests will pass because I will be making a new issue instead of increasing the scope of this issue.


LiamF-FS (approved):
After discussion with Kashike, the scrapers will be moved out of the backend directory and into an etl_pipelines directory.

LiamF-FS (changes requested):
I had meant to request changes, not to approve. My apologies.

LiamF-FS (approved):
This looks good to me! I believe @NathanEne will have more insight regarding the scrapers, however structurally this seems excellent. Cheers

etl_pipelines/utils/etlpipeline_child_classes.py
```diff
@@ -0,0 +1,135 @@
...
+            # __downloaded_data contains the path to the downloaded data if go_through_all_stations is False
+            if not self.go_through_all_stations:
+                self._EtlPipeline__downloaded_data.append({url[0]:data_df})
+
+        # Check if the number of failed downloads is greater than 50% of the total number of downloads if it is, the warnings are promoted to errors
+        if failed_downloads/len(self.source_url) > FAIL_RATIO:
+            logger.error(f"More than 50% of the data was not downloaded, exiting")
+            raise RuntimeError(f"More than 50% of the data was not downloaded. {failed_downloads} out of {len(self.source_url)} failed to download. for {self.name} pipeline")
+
+    def get_station_list(self, station_source = None):
```
NathanEne, line 102: Because we are doing 1 source = 1 pipeline will we be able to have this station source as a class variable and instead of passing it in its just internally used?

etl_pipelines/utils/eltpipeline_class.py
```diff
@@ -0,0 +1,141 @@
...
+        A getter function for the private function __load_data_into_tables.
+
+        Args:   
+            None
+
+        Output: 
+            None
+        """
+        if self.__transformed_data is None:
+            logger.warning("load_data is not implemented yet, exiting")
```
NathanEne, line 51: this warning message seems misleading when compared to the above if statment

etl_pipelines/utils/eltpipeline_class.py
```diff
@@ -0,0 +1,141 @@
...
+        there is some data that is trying to be inserted. If there is not, it will raise an Error.
+
+        Args:
+            None
+
+        Output:
+            None
+        """
+        if self.__transformed_data is None:
+            logger.warning("__load_data_into_tables is not implemented yet, exiting")
```
NathanEne, line 69: This warning is also slightly misleading.

etl_pipelines/utils/eltpipeline_class.py
```diff
@@ -0,0 +1,141 @@
...
+                # Turning dataframe into insertable tuples.
+                records = df.rows()
+
+                # Creating the insert query
+                insert_query = f"INSERT INTO {insert_table_name} ({', '.join(df_schema)}) VALUES %s ON CONFLICT ({', '.join(pkey)}) DO UPDATE SET value = EXCLUDED.value;"
+
+                ### How this connection is got will change once this is in Airflow
+                cursor = db.conn.cursor()
+
+                execute_values(cursor, insert_query, records, page_size=100000)
```
NathanEne, line 109: I like this, execute values is a great starting point. We can definitely look at copy or mogrify if this is too slow but I imagine this will be great

etl_pipelines/utils/eltpipeline_class.py
```diff
@@ -0,0 +1,141 @@
...
+
+        Output: 
+            None
+        """
+        if self.__transformed_data is None:
+            logger.warning("load_data is not implemented yet, exiting")
+            return
+
+        logger.debug(f"Loading data into tables: {", ".join(self.destination_tables.values())}")
+        self.__load_data_into_tables()
```
NathanEne, line 55: if this function is simply a wrapper for `__load_data_into_tables()` we can likely remove it. In the original UML diagram this function was intended to loop through the transformed data and the respective destination tables that they were meant to find their way into and run `__load_data_into_tables()` for each one but if we are not doing that then we should be good to cut it and replace its contents with whats in `__load_data_into_tables()`

etl_pipelines/utils/etlpipeline_child_classes.py
```diff
@@ -0,0 +1,135 @@
...
+
+            logger.debug(f"Downloading data from URL: {url[1]}")
+            # Within function retry, will only retry once.
+            while True:
+                try:
+                    # Stream is True so that we can download directly to memory
+                    response = requests.get(url[1], stream=True, headers=HEADER, timeout=20)
+                except requests.exceptions.RequestException as e:
+                    if self._EtlPipeline__download_num_retries < MAX_NUM_RETRY:
+                        logger.warning(f"Error downloading data from URL: {url[1]}. Retrying...")
```
NathanEne, line 51: We should likely add a small wait between retries to not get ourself timed out on the more sensitive data sources

NathanEne: ~ 5 - 10 seconds is more then enough

etl_pipelines/utils/etlpipeline_child_classes.py
```diff
@@ -0,0 +1,135 @@
...
+                    failed = True
+                    break
+            
+            # If failed flag is True them increment filed_downloads by 1 and move on to next URL
+            if failed:
+                logger.warning(f"The URL {url[1]} failed to download 3 times, moving on to next URL")
+                failed_downloads += 1
+                continue
+            
+            ## This may have to change since not all sources are CSVs
```
NathanEne, line 78: I agree, this is fine for now but we may need to store the raw response then deal with it per scraper in the data transformation step

etl_pipelines/utils/etlpipeline_child_classes.py
```diff
@@ -0,0 +1,135 @@
...
+            if data_df.limit(1).collect().is_empty():
+                logger.warning(f"Downloaded data is empty for URL: {url[1]}. Will not mark as failure but be noted.")
+                continue
+
+            # __downloaded_data contains the path to the downloaded data if go_through_all_stations is False
+            if not self.go_through_all_stations:
+                self._EtlPipeline__downloaded_data.append({url[0]:data_df})
+
+        # Check if the number of failed downloads is greater than 50% of the total number of downloads if it is, the warnings are promoted to errors
+        if failed_downloads/len(self.source_url) > FAIL_RATIO:
```
NathanEne, line 98: `FAIL_RATIO` may need to be defined on a per class basis, some scrapers like the AWT scraper (not included in this project but just an example) are still functioning as expected when they only get 20% of their data but some scrapers like climate stations that don't go down like stream-flow stations do likely want closer to 100%

etl_pipelines/utils/eltpipeline_class.py
```diff
@@ -0,0 +1,141 @@
...
+
+        insert_keys = self.destination_tables.keys()
+
+        # Check that the destination tables have been populated
+        if len(insert_keys) == 0:
+            logger.error(f"The scraper {self.name} did not give any tables to insert to! Exiting")
+            raise RuntimeError(f"The scraper {self.name} did not give any tables to insert to! Exiting")
+
+        # Check that private attribute __transfromded_data has values
+        if not self.__transformed_data:
```
NathanEne, line 80: due to the fact we are expecting this to be a series of polars data frames at this point is this a sufficient check to see if its empty. I know for pandas evaluating `if df:` is always true and you need to use `.is_empty()`

etl_pipelines/scrapers/water/wsc_hydrometric.py
```diff
@@ -0,0 +1,110 @@
...
+import polars as pl
+
+class WscHydrometricPipeline(StationObservationPipeline):
+    def __init__(self):
+        # Initializing attributes in parent class
+        super().__init__(name=WSC_NAME, source_url=[], destination_tables=WSC_DESTINATION_TABLES)
+
+        # Initializing attributes present class
+        self.days = 2
+        self.station_list = super().get_station_list(station_source=WSC_STATION_SOURCE)
```
NathanEne, line 20: do you need to include super here because you are not overriding the `get_station_list` method in this class? Additionally do you anticipate this being a common step for every class? If so we can put that in the `StationObservationPipeline`'s constructor so we dont have to write it every time in here! Also `get_station_list` could likely write the result directly into the `station_list` class variable too avoid having to assign it manually like this. Unless there are cases where you believe we will have to have 2 station lists in the same scraper?

etl_pipelines/scrapers/water/wsc_hydrometric.py
```diff
@@ -0,0 +1,110 @@
...
+        # Note that Once we use airflow this may have to change to a different way of getting the date especially if we want to use
+        # it's backfill or catchup feature.
+        date_now = datetime.now().strftime("%Y%m%d")
+        self.source_url = [["wsc_daily_hydrometric.csv", WSC_URL.format(date_now)]]
+
+        self.end_date = datetime.now(pytz.timezone("UTC"))
+        self.start_date = self.end_date - timedelta(days=self.days)
+        
+
+    def transform_data(self):
```
NathanEne, line 31: This function looks awesome, super sick how readable and quick polars is!

etl_pipelines/scrapers/water/wsc_hydrometric.py
```diff
@@ -0,0 +1,110 @@
...
+            raise RuntimeError(f"Error when applying transformations to the discharge values. Error: {e}")
+
+        # Set the transformed data
+        self._EtlPipeline__transformed_data = {
+            "level": [level_df, ["station_id", "datestamp"]],
+            "discharge": [discharge_df, ["station_id", "datestamp"]]
+        }
+
+
+    def validate_downloaded_data(self):
```
NathanEne, line 109: Re: your `validate_download_data()` question:

I believe `validate_download_data()` was originally included in the UML as a way for us to verify that the data we got back had the right shape ie: contained all expected column names and had data in all of them. Just a quick scan nothing too intense. it is possible we could even put it in the super class if we filled in required columns as a class field.

NathanEne: This is somewhat related to your other question. I am perfectly fine with a hard coded column names we expect. We can do it on a per scraper basis. I am not sure if you forsee any issues with this but we likely want to be notified if the shape of our incoming data changes anyways so allowing the scraper to fail on it and hopefully fixing it manually rather then trying to brute force through it and possibly ending up inserting bad data into the db. If we implement the catchup airflow functionality I don't think missing a run or two will be a big deal. Additionally it is easier to develop and maintain!

NathanEne (commented):
Whole lot of comments! Sorry to give you so much but this is kind of our best time to iron out the structure of things before it gets too set in stone. Let me know if you have any questions always happy to talk.

etl_pipelines/utils/etlpipeline_child_classes.py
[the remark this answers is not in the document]
Kashike-Umemura: I just decided to put the child abstract classes in to one file and the parent in another. I am totally down to split them down in to two separate files and change the dir structure to be as you said. It will look like:
etl_pipelines
- scrapers (contains class structure)
    - specific_implementations (dir name TBD, contains the specifc implementations)

etl_pipelines/utils/constants.py
```diff
@@ -0,0 +1,67 @@
...
+        "https://data-donnees.az.ec.gc.ca/api/file?path=/substances/monitor/national-long-term-water-quality-monitoring-data/peace-athabasca-river-basin-long-term-water-quality-monitoring-data/Water-Qual-Eau-Peace-Athabasca-2000-present.csv",
+        "https://data-donnees.az.ec.gc.ca/api/file?path=/substances/monitor/national-long-term-water-quality-monitoring-data/pacific-coastal-basin-long-term-water-quality-monitoring-data/Water-Qual-Eau-Pacific-Coastal-Cote-Pacifique-2000-present.csv",
+        "https://data-donnees.az.ec.gc.ca/api/file?path=/substances/monitor/national-long-term-water-quality-monitoring-data/okanagan-similkameen-river-basin-long-term-water-quality-monitoring-data/Water-Qual-Eau-Okanagan-Similkameen-2000-present.csv",
+        "https://data-donnees.az.ec.gc.ca/api/file?path=/substances/monitor/national-long-term-water-quality-monitoring-data/lower-mackenzie-river-basin-long-term-water-quality-monitoring-data-canada-s-north/Water-Qual-Eau-Mackenzie-2000-present.csv",
+    ]
```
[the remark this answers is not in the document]
Kashike-Umemura: Yep that makes sense! 
I was originally thinking about importing to populate the attributes because the `__init__` started to get really dense. But I can put them in and anyhow

etl_pipelines/utils/etlpipeline_child_classes.py
Kashike-Umemura: Yes that will be possible. I will change that

etl_pipelines/utils/eltpipeline_class.py
Kashike-Umemura: Fair, but this will be removed once the method is implemented so I think it'll be ok for now

etl_pipelines/utils/eltpipeline_class.py
Kashike-Umemura: Same as above

etl_pipelines/utils/eltpipeline_class.py
Kashike-Umemura: I think it's good enough for most scrapers. the first one I implemented only inserted ~ 2500 rows, so I think execute_values will suffice for most cases

etl_pipelines/utils/eltpipeline_class.py
Kashike-Umemura: Ah I see, I can change it to be an iterator of the __transformed_data, I wasn't 100% sure what it was and assumed it was a caller for the private function.

etl_pipelines/scrapers/water/wsc_hydrometric.py
Kashike-Umemura: I think it doesn't have to use super() since I've been doing it in other places. But I'll double check and change that. 

I don't think there's an issue with `get_station_list` assigning the list to the attribute `station_list` I'll test it out and see if I encounter any issues.

etl_pipelines/utils/eltpipeline_class.py
Kashike-Umemura: Since I'm using a LazyFrame from polars, it's `is_empyt()` doesn't work on it (throws error saying it needs to be materialized first). There are ways to get around it. But I think I'll change the `load_data()` function to iterate through the data that needs to be inserted.

etl_pipelines/scrapers/water/wsc_hydrometric.py
Kashike-Umemura: While writing tests, I found out that the LazyFrame query planner doesn't actually throw any exceptions even if the transformations are illegal. Meaning that the `try except` blocks aren't doing anything. I think I'll remove it and move the `collect()` to the end of the function instead of right before insertion since that allows for checking if it's empty earlier

etl_pipelines/scrapers/water/wsc_hydrometric.py
Kashike-Umemura: Sounds good! I'm going to hardcode the columns for now, unless there are major issues for a specific source. 

Makes sense about the `validate_download_data()` I'll make some some simple checks to make sure that the shape of the data we're getting (mostly column names and dtypes) are what we are expecting

etl_pipelines/utils/constants.py
```diff
@@ -0,0 +1,67 @@
+import logging
+import os
+from dotenv import load_dotenv, find_dotenv
+
+load_dotenv(find_dotenv())
+logger = logging.getLogger('scraper')
+
+FAIL_RATIO = 0.5
+
+DB_URI = f"postgresql://{os.getenv("PGUSER")}:{os.getenv("PGPASS")}@{os.getenv("PGHOST")}:{os.getenv("PGPORT")}/{os.getenv("PGDB")}"
```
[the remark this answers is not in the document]
NathanEne: awesome!

etl_pipelines/utils/constants.py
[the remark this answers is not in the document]
NathanEne: I think importing to populate them can be good, it will give us a single place to see all the links that populate in the constructors.

etl_pipelines/utils/eltpipeline_class.py
NathanEne: sounds good

etl_pipelines/utils/eltpipeline_class.py
NathanEne: whatever works for you honestly, I think structurally they are almost identical. It may make more sense to have it all in a single function so it can all be overrode simultaneously.

etl_pipelines/utils/eltpipeline_class.py
NathanEne: sounds good to me!

etl_pipelines/scrapers/water/wsc_hydrometric.py
NathanEne: Interesting so you can build illegal queries then only get the exception when you attempt to collect it?

etl_pipelines/scrapers/water/wsc_hydrometric.py
NathanEne: Awesome! Sounds good!

etl_pipelines/utils/etlpipeline_child_classes.py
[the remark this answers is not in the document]
NathanEne: would it make sense to do 

- scrapers
  - ETLPipeline.py (superclass)
  - StationObservationPipelines (folder)
    - StationObservationPipeline.py (subclass of etlpipeline)
    - climate (folder with its related scrapers)
    - water (folder with its related scrapers)
  - DataBcPipelines (folder)
    - DataBcPipeline.py (subclass of etlpipeline)
    - licences (folder with its related scrapers)
  - QuarterlyPipelines (folder with its related scrapers as they may need to be quite custom)

etl_pipelines/scrapers/water/wsc_hydrometric.py
Kashike-Umemura: Yeah, ran in to that when I was trying to cause exceptions for testing, but then it just ran completely fine

NathanEne (changes requested):
Requesting changes after discussing with Kashike on slack (mostly as a reminder for him to not merge this)

etl_pipelines/utils/database.py
```diff
@@ -0,0 +1,38 @@
...
+from dotenv import load_dotenv, find_dotenv
+
+load_dotenv(find_dotenv())
+port = os.getenv("PGPORT")
+user = os.getenv("PGUSER")
+password = os.getenv("PGPASS")
+database = os.getenv("PGDB")
+host = os.getenv("PGHOST")
+pgschema_prefix = os.getenv("PGSCHEMA_PREFIX")
+
```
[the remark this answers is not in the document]
Kashike-Umemura: Tried to use the DB_URI constant in the `constants.py` file, but the ThreadedConnectionPool requires each value. Since this will be replaced anyways, going to keep it this way for now

NathanEne (approved):
LGTM, Kashike and I discussed a few changes that will be included in the upcoming PR that adds testing

PR #10, opened 2025-04-16
Bcwat 81 map filters

MaxDouglas1991:
# Description

Update Map Filters to match new mockups.
- Filters moved to LHS
- Active point moves off the map and into top of sidebar
- Added text filter


client/src/components/MapFilters.vue
```diff
@@ -1,80 +1,167 @@
...
+                    :key="button"
+                    v-model="button.value"
+                    :label="button.label"
+                    :color="button.color"
+                    @update:model-value="emit('update-filter', localFilters)"
+                />
+            </div>
+            <div
+                v-if="activePoint"
+                class="selected-point"
```
idaeusice, line 17: I might remove the selected-point class and do the following:
`class="q-pa-md"`
and swap the `div` for `q-card`

idaeusice (approved):
Looks great! Just left a note on a card element as opposed to handling it manually, but works either way!

PR #14, opened 2025-04-17
Bcwat 86 report toc

MaxDouglas1991:
# Description

Create Report Table of Contents
User can click on a section in sidebar to scroll to it
Scrolling in the report highlights whichever section you scroll past


idaeusice (approved):
The streamflow report style has been adversely affected, but we should have some custom styling since that data is represented differently. Looks good.

PR #39, opened 2025-04-17
OKD Deployment, Helm Initialization for App/Airflow

LiamF-FS:
# Description

The goal of this task was to initialize the deployment of the source code via Helm. As there is no access granted to the BCGov openshift as of yet, everything is in reference to the internal registry of our self hosted OKD.

## Types of changes

- Initialized App Deployment
  - Hello World RESTful API, simple NGINX FrontEnd
- Initialized Airflow Deployment
  - Hello World DAG

Turned off Github Actions for the time being, as we do not have any unit tests as of yet.

## Checklist

## Further comments

I am hopeful this acts as a solid foundation for managing our deployments across various environments. The next step is to iron our CI/CD.


airflow/pod_templates/simple_task_template.yaml
```diff
@@ -0,0 +1,88 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: airflow-worker-simple
+  labels:
+    tier: airflow
+    component: worker
+    release: airflow
+  annotations:
+    cluster-autoscaler.kubernetes.io/safe-to-evict: "false"
```
AndersFarrFoundry, line 10: What will this annotation do. Does it mean that if we get under memory pressure then the workers won't be evicted? Could be dangerous for something like the hydat job that is likely lower priority than the pods running the live site. I know this is the defaults and you don't have to change, but probably something to consider?

charts/app/Chart.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v2
+name: quickstart-openshift
+description: Helm Chart for deploying the Client & Backend of the BC Water Tool Consolidation
```
AndersFarrFoundry, line 3: Maybe better to say client + API here for clarity?

charts/app/templates/backend-deployment.yaml
```diff
@@ -0,0 +1,40 @@
...
+      containers:
+        - name: {{ .Values.backend.name }}
+          image: {{ .Values.backend.image }}
+          imagePullPolicy: Always
+          # TODO: Init External Secrets Operator
+          env:
+            - name: FLASK_ENV
+              value: okd
+            - name: CLIENT_URL
+              value: http://bcwatertool.local
```
AndersFarrFoundry, line 29: please change to bcwatertool.fspatial :)

charts/app/templates/configmap.yaml
```diff
@@ -0,0 +1,69 @@
...
+kind: ConfigMap
+metadata:
+  name: "{{ .Values.client.name }}-nginx-conf"
+  namespace: {{ .Values.client.namespace }}
+  labels:
+    app: "{{ .Values.client.name }}"
+    project: {{ .Values.global.project }}
+data:
+  nginx.conf: |
+    user  nginx;
```
AndersFarrFoundry, line 11: Note: I've been removing the user nginx from here as we are not running as nginx user for okd/openshift, but not a huge deal

AndersFarrFoundry (commented):
I've been nitpicking like crazy in here. Minor changes. Great job for the most part I know it was tough! :+1:

airflow/pod_templates/simple_task_template.yaml
LiamF-FS: From the documentation:

```
cluster-autoscaler.kubernetes.io/safe-to-evict 
Type: Annotation

Example: cluster-autoscaler.kubernetes.io/safe-to-evict: "true"

Used on: Pod

When this annotation is set to "true", the cluster autoscaler is allowed to evict a Pod even if other rules would normally prevent that. The cluster autoscaler never evicts Pods that have this annotation explicitly set to "false"; you could set that on an important Pod that you want to keep running. If this annotation is not set then the cluster autoscaler follows its Pod-level behavior.
```

Therefore, I think that you are right in that these pods will not be evicted if another pod with higher priority class requires resources being used by the running pod. Which I feel as though is the expected behaviour...? I think that this could become messy and there could be complications that arise from pods getting evicted.

I feel as though we likely want all pods to have `safe-to-evict` set to false and rather have the workflow that the underlying KubernetesOperator will build an internal queue and allocate resources to pods within that queue based upon their resource requests.

charts/app/Chart.yaml
LiamF-FS: Good catch yeah. Do you think I should rename either the client or backend folder to align better with `client/server` or `frontend/backend`? It feels kind of odd with the syntax we have chosen.

charts/app/templates/backend-deployment.yaml
LiamF-FS: Good catch!

airflow/pod_templates/simple_task_template.yaml
AndersFarrFoundry: Yeah we probably have that on all of them then we're fine. I'm thinking now that this is good for scrapers. I used the priorityClass for the flux scrapers so they wouldn't be evicted over other pods while they were running. I think that if an APi is evicted, it'll just be scheduled in another spot so it's probably fine. Sorry to annoy you yeah this is good.

charts/app/Chart.yaml
AndersFarrFoundry: I think client + api works? That is like our other projects

AndersFarrFoundry (approved):
lgtm thank you!

airflow/pod_templates/simple_task_template.yaml
LiamF-FS: No problem at all - it is super good to be mindful of these. Cheers!

PR #42, opened 2025-04-17
Bcwat 20 annual hydrology

MaxDouglas1991:
Add Annual Hydrology Section to watershed report
Opening the report now requires the user to click on the map. The location clicked is used in annual hydrology map

Add Allocations By Industry Section


idaeusice (approved):
Nice changes, things look good!

Issue #41, opened 2025-04-17
Make Initial DAG for WSC Hydro Scraper


Kashike-Umemura:
Once the WscHydrometricPipeline class has been implemented, it will be ready to be scraped using AirFlow.

The goal of this issue is to create the DAG for the local deployment of AirFlow, and ensure that it works.

For this to be completed, there will have to be some changes in the class itself, mainly the following:
- Change anything that uses the `datetime.now()` function to use the logical date instead
- Anything that uses a db connection uses the Postgres connection hook

PR #59, opened 2025-04-22
Bcwat 41 wsc hydro pipeline dag

Kashike-Umemura:
# Description

- Completed Issue #41 in this PR.
- The `etl_pipelines/` directory was moved in to the `airflow/` directory so that the Docker container can have access to them when it is built.
- Fixed any issues that happened with directory move
- Added `__init__.py` files to make the `etl_pipeline/` directories in to a package.
- Created DAG that allows for logging and automatic running of the WSC hydrometric scrapers

## Types of changes

## Checklist


airflow/dags/wsc_hydro_dag.py
```diff
@@ -0,0 +1,46 @@
+import pendulum
```
LiamF-FS, line 1: I do not see pendulum within either requirements.txt

Furthermore, I think we should just have the single requirements.txt within airflow root folder.

This is because when dockerizing, we should just need to perform the single pip install, rather than both - I understand that maybe the nested requirements file is there for the test suite only, but I still think they could activate virtual env, and then install before going into pipelines directory and running tests.

airflow/Dockerfile
```diff
@@ -5,8 +5,11 @@ USER root
 COPY dags/ /opt/airflow/dags/
 COPY plugins/ /opt/airflow/plugins/
 COPY pod_templates/ /opt/airflow/pod_templates/
+COPY etl_pipelines/ /opt/airflow/etl_pipelines/
 
-RUN chown -R airflow: /opt/airflow/dags /opt/airflow/plugins
+RUN chown -R airflow: /opt/airflow/dags /opt/airflow/plugins /opt/airflow/etl_pipelines
+
+ENV PYTHONPATH = "/opt/airflow/etl_pipelines:${PYTHONPATH}"
```
LiamF-FS, line 12: Is this required?

LiamF-FS (commented):
A few comments being addressed - all tests passing, just going to manually build and deploy on OKD and see how it goes!

airflow/dags/wsc_hydro_dag.py
Kashike-Umemura: Removed the `requirements.txt` in `etl_pipelines/` and made sure that pendulum is in the one for `airflow/`

airflow/Dockerfile
Kashike-Umemura: I believe so, default python/airflow only adds `airflow/dags/`, `airflow/plugin/`, and `airflow/config/` to the `PYTHONPATH` variable.

Adding that makes sure Python checks that directory for modules that I refer in the files as well.

LiamF-FS (approved):
All looks good! 

Moving forwards, create more pod_templates with differing resource requests/limits.

Also it may be time to retire the hello_world_dag, as we have this one working.

Great Job!!

PR #60, opened 2025-04-23
21 recreate watershed report monthly hydrology

MaxDouglas1991:
# Description

Recreate Monthly Hydrology Section including:

- Stacked Bar Chart with legend
- table

for both the query and downstream watersheds


idaeusice (approved):
Looks OK. The hover effect on the chart is disappearing if you hover over the dotted (MAD) lines. This can be avoided by adding a style property for `pointer-events: none` on the class associated with the MAD lines.

PR #73, opened 2025-05-01
Bcwat 52 monthly mean flow

idaeusice:
# Description
Including the Monthly Mean Flow table in the streamflow report popup. This includes a table with gradient shading based on the values for each row for the indicated years, as well as for minimum, maximum, and mean rows. 

- introduced a new table
- added color gradient handling within the table
- added a fixture data file for monthly mean flow data

## Types of changes

## Checklist

## Further comments
This table only uses fixture data for now. In the future, changes will be made to use an API call. The `getTableData` function has been set up to serve this purpose in the future, so all that needs to be done is update the function to await the API response.

### DIFF
--- a/client/src/components/streamflow/MonthlyMeanFlow.vue
+++ b/client/src/components/streamflow/MonthlyMeanFlow.vue
@@ -59,13 +59,14 @@ const setTableData = () => {
     // set the rows
     tableRows.value = [];
 
+
+    const foundVars = [
+        { name: 'Mean', type: 'avg', found: false }, 
+        { name: 'Maximum', type: 'max', found: false }, 
+        { name: 'Minimum', type: 'min', found: false }, 
+    ];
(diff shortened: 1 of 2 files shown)


client/src/components/streamflow/MonthlyMeanFlow.vue
```diff
@@ -1,7 +1,128 @@
...
+            field: month,
+        });
+    });
+
+    // set the rows
+    tableRows.value = [];
+
+    // populate rows with mean, max, min data
+    tableData.value.monthly_mean_flow.current.forEach(el => {
+        const foundVars = [
```
MaxDouglas1991, line 64: This can be moved outside the forEach to avoid redeclaring it repeatedly

MaxDouglas1991 (approved):
LGTM. I left a comment about a minor optimisation

PR #75, opened 2025-05-01
24 recreate watershed report hydrologic variability

MaxDouglas1991:
# Description

Recreate Hydrologic Variability section in watershed report
This component includes:

- A map with the query watershed and 8 other similar watersheds to compare against
- A table for the query watershed and its contemporaries including d3 charts for rain, snow and temperature
- A grouped d3 bar chart
- A table including the data from the bar chart


idaeusice (approved):
Looks great! Nice work

PR #77, opened 2025-05-01
Bcwat 68 new station flowworks

Kashike-Umemura:
# Description

Overall, the changes in this PR are the following:
- FlowWorks scraper will check for new stations that are not in the database, and if any are found, they will be inserted after the following checks:
    - The station is
- The station will be by default have the `scrape` flag as `True`, The only exception is when there is no variables of interest associated with the station. Then the station will be inserted, but with the `scrape` flag set to `False`
    - This can be reviewed to not insert at all if we want to.
- Associated metadata for the stations will be inserted to the following tables:
    - `station_variable`, `station_project_id`, `station_year`, and `station_type_id`
    - The data that is supposed to go into `station_region` data is calculated by an trigger on insert of the new stations.
- At the end of the scraper, the stations that got new data is checked to see if that year exists in the `station_year` table.

## Types of changes

## Checklist

## Further comments

The function to check if the new stations are in BC, and checking if the current year is in the `station_year` table is new and will have to be implemented for the `wsc_hydrometric`, and `env_hydro` scraper still.

Also, this is currently ran daily after each scraper run. Although I feel like this is a bit overkill, I feel like it's also a bit necessary since some stations might not get data as frequently.


Kashike-Umemura:
The add new stations method is getting an overhaul, It does work currently but I'm making it better. Will be a different PR so don't worry about going into detail on this one

NathanEne (approved):
LGTM

PR #116, opened 2025-05-06
99 climate report   precipitation

MaxDouglas1991:
Recreate the Climate page's precipitation chart using fixture data


idaeusice (approved):
LGTM!

PR #120, opened 2025-05-06
Adapt climate chart

MaxDouglas1991:
Turn the Precipitation chart component into a resuable component that can be used for precipitation, temperature, snow water equivalent and snow of ground charts in the Climate report


idaeusice:
The old chart includes the current date as a little always-present popup on the chart which does not appear to be included here. It should just list today's date at the x-position of that line. 

Screenshot_20250507_093613

idaeusice:
We should remove the .domain elements - this is done in a number of d3 example charts and will prevent the black border around the chart, which is also not present on the example images.

idaeusice (approved):
Looks good!

PR #121, opened 2025-05-07
added dotenv and try block

Kashike-Umemura:
Minor change to make sure that dot env can load


LiamF-FS (approved):
Awesome - Will test changes after this goes. Hoping this resolves the issues!

PR #122, opened 2025-05-07
changed env-var-name

Kashike-Umemura:
minor change to fix airflow


LiamF-FS (approved):
Debugging! thanks kash

PR #124, opened 2025-05-07
Clean Up Query

LiamF-FS:
Clean Up Query for Flowworks Scraper


Kashike-Umemura (approved):
Thank you! I'll double check that this is sql injection proof in a bit

PR #126, opened 2025-05-08
Bcwat 80 asp scraper

### LINKED ISSUE #80: Implement `asp` scraper
# Purpose

Just like the other scrapers, implement the `asp` scraper. See the other scrapers completed before this to see the general structure.

## Completion Criteria

- [x] Download Function
- [x] Transform Function

Kashike-Umemura:
# Description

ASP Scraper has been implemented! Following are note able changes that were introduced to other parts of the scrapers:
#### New class attribute `overrideable_dtype` has been added to `StationObservationPipeline` class
This was added because `asp` introduced a dtype schema that would have been impossible/impractical to define. It is Date as one column, and each unique stations' `original_id` as it's other columns. This would have made the dtype schema ~110 entries long. This allows me to skip the `schema_override` argument for `pl.scan_csv`.

I am open to discussion on this change

#### Reformatting some parts of other scrapers
Noticed that some of the scrapers had it's Polar's `select`, and `with_columns` statements in one long line. This made it hard to read so reformatted it to be on multiple lines. Increases lines but improves readability.

#### Moved `swe` to have it's mean calculated, max for `pc`, and sum for `rainfall` in the FlowWorks scraper.
While working on the ASP scraper, I noticed that I did those three parameters wrong so I fixed them up. I made sure that they still ran without issue to the end after I changed them.

#### Standardized `date_now` to UTC, and `start_date`/`end_date` to America/Vancouver.
I noticed that I was flip-flopping the two around and it was becoming inconsistent. So fixed that

## Types of changes

## Checklist

## Further comments

I have noticed that the `bcwat_obs.climate_precip_amount`. `bcwat_obs.climate_msp` table will be getting multiple variable id's that are not really related to each other. I say "not related" because the `bcwat_obs.climate_temperature` table gets min, max, and mean temperatures, they are still talking about the same variable, just different methods of calculating the daily value. 

But the other two will be getting values such as `lwe_precipitation_amount`, `precipitation_amont`, `precipitation_between_0600_1800` for the former, and `msp_snow_depth`, `msp_percent_normal`,`msp_swe` for the latter. I think that instead of having them as the same table it might be better to either split them. or consolidate all snow related values in to one table, precipitation related values to another, and so on. I know that this would be diverging from the data model that Nathan and I created, but I wanted to bring it up to see how people would feel.

Thanks!

(Closes #80)


NathanEne (approved):
LGTM

PR #135, opened 2025-05-12
Bcwat 81 drive bc scraper

### LINKED ISSUE #81: Implement `drive_bc` scraper
# Purpose

Just like the other scrapers, implement the `drive_bc` scraper. See the other scrapers completed before this to see the general structure.

## Completion Criteria

- [ ] Download Function
- [ ] Transform Function

### NOTE
Since some of these use APIs to download the data, it might be a good idea to make another class, or implement API specific methods to `StationObservationPipeline` this may increase the tickets.

Kashike-Umemura:
# Description

DriveBC scraper implementation!
Was relatively simple, only the transformation function, and since this is the scraper that runs hourly, the data just gets inserted directly. Will make the aggregation function once it has at least one days worth of data scraped so that I can test.

Other than that, the following changes were made:
- Added small snippet to the `StationObservationPipeline.download_data` function that removes leading and trailing white spaces to column names.
- Added some more conditionals for the `drivebc - moti` data since it is in JSON. If there are more scrapers that take json strings in then I will make a proper flag.
- Fixed typo `stationd_data` in all scraper files
- Moved the date initialization to the `StationObservationPipeline` class since it was being done in all places. Even if it isn't used, I think this is better.
- 
## Types of changes

## Checklist

## Further comments

closes #81


airflow/etl_pipelines/utils/constants.py
```diff
@@ -229,42 +229,74 @@
...
-        " Elev. meters": pl.Int64,
-        " Date of Survey": pl.String,
-        " Snow Depth cm": pl.Int64,
-        " Water Equiv. mm": pl.Int64,
-        " Survey Code": pl.String,
-        " Snow Line Elev. m": pl.Int64,
-        " Density %": pl.Int64,
-        " Survey Period": pl.String
+        "Number": pl.String,
+        "Elev. metres": pl.Int64,
```
Kashike-Umemura, line 244: To be honest, I don't know how this didn't trigger the validation error. This old version is currently in the dev airflow env and runs fine. But removing the whitespace from the string some how caused it to trigger?
Or maybe the data file that we get switches over sometimes?
Will check

NathanEne: Let me know what you find, strange this is not triggering the error

airflow/etl_pipelines/scrapers/StationObservationPipeline/StationObservationPipeline.py
```diff
@@ -95,16 +128,30 @@ def download_data(self):
...
             try:
                 logger.debug('Loading data into LazyFrame')
                 response.raw.decode_content = True
-                if self.go_through_all_stations:
+                # This is for the DriveBC scraper since it get's a JSON string. If there more scrapers that returns a JSON string then I will make this a proper
+                # flag. But for now this is the way.
+                # There are also stations with "" as column values. JSON loads does not play well with that, which confuses the pl.LazyFrame constructor. So replace all instances of ""
+                # with "No Data Reported"
+                if self.station_source == "moti":
+                    data_df = pl.LazyFrame([row["station"] for row in json.loads(response.text.replace('""', '"No Data Reported"'))], schema_overrides=self.expected_dtype["drive_bc"])
```
NathanEne, line 141: this if chain is gettign a bit long, do you think it would be helpful to extract this logic into a private function that gets called that each scaraper can either implement or use an implementation from a super class?

NathanEne (approved):
LGTM, just a structural qeustion

airflow/etl_pipelines/utils/constants.py
Kashike-Umemura: Sounds good :)

airflow/etl_pipelines/scrapers/StationObservationPipeline/StationObservationPipeline.py
Kashike-Umemura: I think that would be helpful! Since I'm working on another scraper and I already need to add another if statement on the chain. I'll switch over to that method in this branch

NathanEne (approved):
LGTM

PR #136, opened 2025-05-12
bcwat-100-surface-water-quality

idaeusice:
# Description

Added the page and functionality for the surface water charts. The supplied mockup indicated a number of columns for each parameter for the data points that would be present in the charts, however this is impossible as each chart is not guaranteed to contain the same number of points. I chose to include these columns as one column, which is then broken down in each parameter (row). Open to suggestions on other/better ways to represent this data. 

## Types of changes

## Checklist

## Further comments
Referencing fixture data for the data included in this feature.


client/src/components/surfacewater/SurfaceWater.vue
```diff
@@ -0,0 +1,242 @@
...
+import { highlightLayer, pointLayer } from "@/constants/mapLayers.js";
+import SurfaceWaterQualityReport from "@/components/surfacewater/SurfaceWaterQualityReport.vue";
+import { ref } from 'vue';
+
+const map = ref();
+const activePoint = ref();
+const features = ref([]);
+const pointsLoading = ref(false);
+const reportOpen = ref(false);
+const streamflowFilters = ref({
```
MaxDouglas1991, line 38: These variable names should match the component they're being used in

MaxDouglas1991: This object is a good candidate for constants.js as well

MaxDouglas1991 (approved):

PR #138, opened 2025-05-14
Bcwat 82 ec xml scraper

### LINKED ISSUE #82: Implement `ec_xml` scraper
# Purpose

Just like the other scrapers, implement the `ec_xml` scraper. See the other scrapers completed before this to see the general structure.

## Completion Criteria

- [ ] Download Function
- [x] Transform Function

### NOTE
This scraper has to go through an xml file for scraping, so it may be a bit weird.

Kashike-Umemura:
# Description

Necessary components for the EC XML scraper:
- DAG for Airflow (Pretty simple and same as before)
- Transform method, as well as a new `make_polars_lazyframe` method. More on that later.
-  Added more constants to the constants file.
- In the last [PR](#135) the part that initialized the polars lazyframe in `StationObservationPipeline` was changed into a private method that was called. This change was made because the string of `if elif ... else` statements were getting long. The private method is overridden for this scraper since xml files aren't natively supported by polars.

Other changes:
- When loading data in to the database, after it checks if the insert DataFrame is empty or not in the `EtlPipeline.load_data()` method, instead of throwing an error if the DataFrame is empty, it will log an warning and email. 
    - This was chosen over throwing an error and exiting the scraper run because it might be expected to happen, and not necessary an error. I could make it so that the transform method only adds DataFrames that have data in them to be sent for insertion as well.
- I turned on "Remove tailing whitespaces on save" in vscode. I'm sorry for all the whitespaces I've been leaving behind.

## Types of changes

## Checklist

## Further comments

This will close #82


NathanEne (approved):
LGTM

PR #140, opened 2025-05-15
Bcwat 87 weather farm prd scraper

### LINKED ISSUE #82: Implement `ec_xml` scraper

Kashike-Umemura:
# Description

Implementation of Weather Farm PRD scraper:
- This data source serves raw JSON from the request. But not all stations have the columns that we are scraping. So this was dealt with by coalescing the columns with None values as default. 
- The scraper also didn't use the `original_id` of the stations at all. They used an UUID that is stored in an JSON object in the database. So I needed to add a new column to the `bcwat_obs.station` table with said data.
- Otherwise this scraper is basically the same as the others.

Other Changes:
- Added a new method to the class `StationObservationPipeline` called `check_number_of_stations_scraped`
    - **What:** This method basically checks the ratio of unique `station_id`s in the transformed_data and the number of stations that were supposed to be scraped. If the ratio is less than the `<scraper>_MIN_RATIO` value, it will send an email and log an warning. The `<scraper>_MIN_RATIO` is unique to each scraper, since then we get to control when the email should be sent (ie. MSP scraper almost never returns data during the summer, so we can set it's expected ratio to be 0). For now I have set it all to be 0.5, and custom values can be discussed later. To accommodate this change, all DAGs all call the `check_number_of_stations_scraped` method after the `transform_data` step.
    - **Why:** The reason this was implemented is because the Weather Farm PRD URL returns status code 200 most of the times. But sometimes it's filled with NODATA values. This function will allow us to make sure that the number of stations inserted is correct if the number of failed downloads don't catch the lack of station 
- More white space fixes (sorry)
 
## Types of changes

## Checklist

## Further comments

I've also thought about moving the `get_and_insert_new_station()` function out of the transform step and make it's own step (ie. putting the method call in the DAG). Any thoughts?

Closes #82


Kashike-Umemura:
The PR #143 that got merged in is documentation for the `backend/database_initialization/` directory. If you want to take a look please do, but Liam has approved the PR so you can skim, or skip over it.

NathanEne:
Would you be able to store this uuid as the original id for these or is the data too complicated and an additional column (or table) is truly needed

backend/README.md
```diff
@@ -81,3 +82,7 @@ pytest tests/test_hello_world.py
 ```bash
 pytest tests/test_hello_world.py::test_hello_world
 ```
+
+## Database
+
+Please see the README in the `database_initialization` for information on what is in that directory
```
NathanEne, line 88: you can actually use a realative file path to do a clickable link here if you would like!

NathanEne (approved):
LGTM (just one question)

backend/README.md
Kashike-Umemura: Mmm that's cool, I'll give it a go!

Kashike-Umemura:
> Would you be able to store this uuid as the original id for these or is the data too complicated and an additional column (or table) is truly needed

This one isn't too complicated (just a uuid) but there are other stations from other networks where the data in it is a bit more complicated. There is no real structure to it, but I think putting it in to a different table would make more sense. Thoughts?

NathanEne:
> This one isn't too complicated (just a uuid) but there are other stations from other networks where the data in it is a bit more complicated. There is no real structure to it, but I think putting it in to a different table would make more sense. Thoughts?

A different table will certainly be better/cleaner in my opinion

Kashike-Umemura:
Actually. I've deemed it unnecessary and will remove this column and make some changes

NathanEne:
awesome that's probably better to not add extra columns only some gauges will populate when possible

PR #142, opened 2025-05-15
bcwat-113-add-map-search

idaeusice:
# Description

This feature adds a search bar with various options to each page with a map. 

## Types of changes

- Added search bar component
- Added search input
- Added search type dropdown selector
- Added constants for Streamflow, Watershed, Climate, Surface Water, and Ground Water that define the searchable types for each page. 

## Checklist

## Further comments
Some of search options are specific to certain pages. For instance, some pages don't have station names, so this option is omitted. Further search options can be added easily as needed by updating the "...SearchTypes" constants in each component.

### DIFF
--- a/client/src/components/MapSearch.vue
+++ b/client/src/components/MapSearch.vue
@@ -2,9 +2,10 @@
     <div class="search-bar-container">
         <q-select 
             :model-value="searchType"
-            :options="props.searchOptions"
+            :options="allSearchOptions"
             map-options
             emit-value
+            label="Search Type"
             bg-color="white"
             dense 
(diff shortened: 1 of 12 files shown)


idaeusice:
I introduced an experimental change to the 'bus stops' on the watershed report, as it was not drawing lines between 'stops', but I don't think it needs to be included here.

MaxDouglas1991:
> I introduced an experimental change to the 'bus stops' on the watershed report, as it was not drawing lines between 'stops', but I don't think it needs to be included here.

I like it

MaxDouglas1991 (approved):

PR #146, opened 2025-05-16
125 climate page   manual snow survey

MaxDouglas1991:
Recreate the Manual Snow Survey chart from KWT using fixture data


idaeusice:
When I include historical lines, the data isn't appearing in the tooltip - also the dotted line is not aligned with the hovered data. 
image

idaeusice (approved):
Looks good!

PR #152, opened 2025-05-16
151 climate report chart legend

MaxDouglas1991:
Incorporate chart legend into climate chart svg so it is included in the png download.


idaeusice (approved):
Looks good!

Issue #90, opened 2025-05-02
Implement `water_rights_applications_public` DataBC scraper


Kashike-Umemura:
# Purpose

Just like the other scrapers, implement the `water_rights_applications_public` scraper. See the other scrapers completed before this to see the general structure.

## Completion Criteria

- [ ] Download Function
- [ ] Transform Function

### NOTE
Since this a scraper for DataBC it may be more difficult than expected. Take Caution

PR #180, opened 2025-06-02
Bcwat 91 water rights licences public scraper

### LINKED ISSUE #91: Implement `water_rights_licences_public` DataBC scraper
# Purpose

Just like the other scrapers, implement the `water_rights_licenses_public` scraper. See the other scrapers completed before this to see the general structure.

## Completion Criteria

- [ ] Download Function
- [ ] Transform Function

### NOTE
Since this a scraper for DataBC it may be more difficult than expected. Take Caution

Kashike-Umemura:
# Description

Implemented the Water Rights Licences Public scraper:
- Tranform function has been implemented
- As per #171 changes were made so that the truncate will always happen in the same db.commit as the insertion
    - This was done by adding a new flag `truncate` to the function `_load_data_into_tables`  so that it can truncate the table
 before inserting.
    - 2minute sleep before retries were added
- New entries in the `etl_pipelines/utils/constants.py` file
- Changed how the `bcwat_lic.bc_wls_wrl_wra` data was being queried from the old database. Instead of collecting all the data in prod, it will collect all the data from staging.
- Three new tables were added:
    - `bcwat_lic.licence_bc_app_land`
    - `bcwat_lic.bc_water_rights_licences_public`
    - `bcwat_lic.bc_water_rights_applications_public`
    The first table is used for filtering some licences out of the DataBC layer. The latter two will be used as temporary tables that will store some licence data. This was made because the two data sources needs to be populated before some values are calculated.

## Types of changes

## Checklist

## Further comments

This will close #91

The DAG for this scraper will be created after #90 is completed. This is because the data from these two will be in the same time. I thought about splitting them, but there is some calculation that will require to join the two separate tables any ways. So I figured that keeping them together, and putting both scrapers in to one Dag will be the easiest way to deal with this.


NathanEne (approved):
LGTM, Some small comments below and conversation about a change already in progress with error handling in person but nothing blocking

PR #193, opened 2025-06-10
RM deploy on PR Open

LiamF-FS:
# Description

Remove Auto Deploy on PR Creation

## Types of changes

Will be turned on in the future, however we must initialized a dedicated dev database initialized, rather than the DB per PR Open


AndersFarrFoundry (approved):
They overcooked us man! Simple solution for now

PR #195, opened 2025-06-10
bcwat-156-streamflow-unit-testing

idaeusice:
# Description
- Added various component/unit tests for streamflow-specific components. In some cases, there is duplicate testing,  as there is additional coverage in the E2E test suite. 
- Updated a number of components with cypress-specific test targeting properties.
- Updated SevenDayFlow and Stage charts to use the common ReportChart component, both sharing the seven day flow fixture data, as both represent data similarly. 
- Updated various components to use the newly-formed json schema and updated some fixture data with this in mind. Ideally, this will prevent having to reformat when the API is added and response data being returned. 

## Types of changes

## Checklist


MaxDouglas1991 (approved):
LGTM

PR #199, opened 2025-06-16
Major data import script update

### LINKED ISSUE #192: Complete script that will populate ALL DATA
# Purpose

Currently, in the `/backend/database_initialization/` directory, the script that populates part of the data exists in the database.
The only data that is in there is the data that is required to get the scrapers to work. But for the frontend to work, it must be populated completely. This issue is to complete the task of finishing the script!

## Completion Criteria

This issue is considered complete when one runs the script to transfer the data, it completely populates the `bcwat-dev` database with all the required data. (ie no empty tables)

Kashike-Umemura:
# Description

The directory `database_initialization` has been updated to import all of the data. This script will have to change a bit in the future so that it is capable of uploading all the data to S3, so that the BC Gov's database can access the data to be redumped in to there.

The main changes include:
- Documentation update
- Changed from client side cursor to server side cursor. The client side cursor was causing all the data to be loading. Which caused multiple OOMs for me.
- Moved the function to actually join the `bcwat_obs.station` table on to the incoming data into the while loop. This was done because when there are more rows to be fetched, if it was done outside of the while loop, the join would not happen.
- Completed the `constants.py` file
- Removed `bcwat_lic.hypsometric_elevation_rollup`, which was added as the `elevs` column to the table `bcwat_ws.fund_rollup_report`
- Updated Flyway migration `.sql` files.
- JSONified a lot of tables which do not get any transformations done on them

## Types of changes

## Checklist

## Further comments
Closes #192


backend/database_initialization/all_data_transfer.py
```diff
@@ -94,52 +118,78 @@ def populate_all_tables(to_conn, insert_dict):
...
             from_conn.rollback()
             from_conn.close()
-            raise RuntimeError
+            raise RuntimeError(f"Something went wrong inserting the large tables!")
 
         to_conn.commit()
+        to_cur.close()
+        from_cur.close()
+        to_conn.close()
+        from_conn.close()
```
NathanEne, line 182: All this closing seems like it could be combined into a finally case

NathanEne (approved):
LGTM, it is still unclear how this data will get loaded into the database. Do we want to wait to merge this until it is done or do we want to get this in now and change it as future work?

backend/database_initialization/all_data_transfer.py
Kashike-Umemura: mm that is a good point, will make the change. Ty!

Kashike-Umemura:
> LGTM, it is still unclear how this data will get loaded into the database. Do we want to wait to merge this until it is done or do we want to get this in now and change it as future work?

I think it'll be fine to make it future work, The change shouldn't be too large since it'll be writing to a different format in the S3 bucket.

Issue #207, opened 2025-06-20
Update MOTI DriveBC API to use new API route


Kashike-Umemura:
The DriveBC API has been unresponsive since June 18th. An email to the ministry of transportation was sent to inquire about it. The following are their response:

> Hello Kashike,
> 
> Thank you for your enquiry regarding availability of Ministry weather data.
> 
> For the time being you can connect through the legacy site at [legacy.drivebc.ca/api/weather/observations?format=json](https://legacy.drivebc.ca/api/weather/observations?format=json) . Please anticipate this to end without notice within a few months.
> 
> Commercial users are encouraged instead to use the API gateway at: [API Services Portal | API Directory | SAWSx API Services](https://api.gov.bc.ca/devportal/api-directory/3285?preview=false)
> 
> I hope this helps.
> 

I have already updated the scrapers to use the legacy API route as a temporary fix. But this will be decommissioned as well, so it will need to be updated again to use the new API gateway.

Please investigate and implement the API gateway method.

Kashike-Umemura:
Marking this as blocked until they accept my API access request

PR #208, opened 2025-06-20
Move `network_id` and `type_id` into the table `bcwat_obs.station`

### LINKED ISSUE #204: Move `network_id` to be part of the `bcwat_obs.station` table instead of being it's own table
Station's require their `network_id`'s to be in the same table. this is because if don't do this, there will be no distinction between whether the station is a climate station or not.

Kashike-Umemura:
# Description

While writing the `quarterly_hydat` scraper, ran in to a issue where the stations already exist in the database under a different network_id. I dealt with this originally by just having a `station_network_id` table which mapped the stations to multiple `network_id`s. The issue is that if one network has only climate data, and the other has hydrometric data, there is no way to tell which network provides which, causing issues during insertion of new stations, as well as when querying for the API.
To make a further distinction between climate, and hydrometric stations/networks, the following changes have been made:

- Removed `bcwat_obs.station_network_id` table, added a `network_id` column to the `bcwat_obs.station` table
- Removed `bcwat_obs.station_type_id` table, added a `type_id` column to the `bcwat_obs.station` table
- Removed the functionality to add new `network_id`s to stations when they were already in the database with a different `network_id`. Now it inserts a new station with a different `network_id` into the `bcwat_obs.station` table
- Removed/Updated associated queries in the `backend/database_initialization` directory
- Updated associated documentation

In addition, this PR contains the minor change in the URL for the DriveBC source since they depricated the original one. This new URL is the legacy URL, and will be decommissioned in the future, so will have to be changed to the API Gateway, which already has an associated issue #207 

## Types of changes

## Checklist

## Further 

Closes #204


LiamF-FS (approved):
All of these changes look good to me, and make sense. I appreciate including the justifications, the decision to port `network_id` and `type_id` into the stations table makes sense to me.

There is 0 issue with tearing down and rebuilding the database with these new changes. 

I imagine some work will need to be done in the future to ensure that all of the data-model changes exist within the flyway migrations (I believe you previously mentioned some were missing) - but that will be visited in the future.

Great job!

PR #210, opened 2025-06-24
Quarterly Hydat Import Scraper 

Kashike-Umemura:
# Description

Quarterly Hydat Import has been implemented. All aspects of this scraper is quite different since we need to access a sqlite3 database to import the data. The following are what got implemented:
- Hydat Dag that uses `heavy_task_template.yaml` which has:
    - 4Gb of Memory allocated to it. Might have to be increased but we will see
    - 3Gb VolumeMount allocated for downloading and unzipping the Hydat DB.
- The scraper has been scheduled to run at 00:00 on the 30th of: Jan, Apr, Jul, and Oct, of every year. This is because a new version of Hydat get's published every quarter. But not exactly on the regular calendar year quarter, so they might be publishing every quarter of a water year.
- Made `wsc_hydrometric` class constructor accept a `days` argument so that we can easily run it for 365 days worth of data.
- Added `bcwat_obs.symbol` for the Hydat data.

Other Changes:
- Changing the Trigger on the `bcwat_obs.station` table to insert `project_id` to the `bcwat_obs.station_project_id` table instead of assuming which project they're in. No idea if we'll use this table.
- ETC.

## Types of changes

## Checklist

## Further comments

The `DAG` still needs to be tested since my local machine can't access the PVC. So merging this PR and I will run the scraper and keep an eye on it to make sure that it'll work.

### DIFF
--- a/airflow/pod_templates/heavy_task_template.yaml
+++ b/airflow/pod_templates/heavy_task_template.yaml
@@ -99,3 +99,12 @@ spec:
     - name: data
       persistentVolumeClaim:
         claimName: airflow-data
+  affinity:
+    nodeAffinity:
+      requiredDuringSchedulingIgnoredDuringExecution:
+        nodeSelectorTerms:
+          - matchExpressions:
+              - key: polars-unsafe # OKD worker0 (should not impact BC Government Openshift)
+                operator: NotIn
+                values:
+                  - "true"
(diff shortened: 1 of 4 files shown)


migrations/sql/V1.0.1__bcwat_obs_erd_diagram.sql
```diff
@@ -300,6 +304,13 @@ CREATE TABLE "bcwat_obs"."station_water_quality_parameter" (
     PRIMARY KEY ("station_id", "parameter_id")
 );
 
+CREATE_TABLE "bcwat_obs"."symbol" (
```
LiamF-FS, line 307: CREATE TABLE

PR #211, opened 2025-06-24
Up API Resources, initialize Node Affinity for Worker Pod Templates

LiamF-FS:
# Description

Mitigate against API OOMing (think its OOMing on load of that massive JSON fixture)
Prevent Pods from getting scheduled on worker0 (not polars compatible)

## Types of changes

## Further comments

To test this out, we gotta merge in the quarterly hydat branch, and trigger DAG and see what happens!

### DIFF
--- a/airflow/pod_templates/medium_task_template.yaml
+++ b/airflow/pod_templates/medium_task_template.yaml
@@ -98,7 +98,7 @@ spec:
       requiredDuringSchedulingIgnoredDuringExecution:
         nodeSelectorTerms:
           - matchExpressions:
-              - key: polars-unsafe # OKD worker0 (should not impact Openshift)
+              - key: polars-unsafe # OKD worker0 (should not impact BC Government Openshift)
                 operator: NotIn
                 values:
                   - "true"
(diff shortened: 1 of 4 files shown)


airflow/pod_templates/medium_task_template.yaml
```diff
@@ -93,3 +93,12 @@ spec:
     - name: config
       configMap:
         name: airflow-config
+  affinity:
+    nodeAffinity:
+      requiredDuringSchedulingIgnoredDuringExecution:
+        nodeSelectorTerms:
+          - matchExpressions:
+              - key: polars-unsafe # OKD worker0 (should not impact Openshift)
```
AndersFarrFoundry, line 101:
```suggestion
              - key: polars-unsafe # OKD worker0 (should not impact BC Government Openshift)
```
Maybe just to make it even more clear

AndersFarrFoundry (approved):
lgtm, thank you for the discussion about these

airflow/pod_templates/medium_task_template.yaml
LiamF-FS: Good shout - I will do that.

PR #235, opened 2025-07-10
bcwat-222-api-fe

idaeusice:
# Description
- Updated all api.js functions to make actual API calls
- Various style updates and improvments
- Various bug fixes
- Updated component tests to work with the API changes by adding intercepts to stations and reports endpoint calls
- Modified charts and tables to work with the new API data response structures
- Updated some of the fixture content to reflect the new changes and work as expected
- Fixed some issues around the search functionality to allow user to search map by station ID or name where applicable

## Types of changes

## Checklist

## Further comments
There is likely much more that will need to be done, but this is a good first pass for the Front End portion of this task.


client/cypress/component/climate/ClimateReport.cy.js
```diff
@@ -1,6 +1,6 @@
 import ClimateReport from "@/components/climate/ClimateReport.vue";
-import activePointClimate from '@/constants/activePointClimate.json';
-import climateReport from '@/constants/climateReport.json';
+import activePointClimate from '../../fixtures/activePointClimate.json';
+import climateReport from '../../fixtures/climateReport.json';
```
LiamF-FS, line 3: I prefer seeing us reference things with `@/` instead of `../../` - is that possible for these cases?

LiamF-FS (approved):
I think this is a brilliant start to the front end for serving live data. Bulk of line changes are Fixture data for tests, so not as daunting as I thought.

Left a few comments regarding syntax for importing fixtures, and certain if chains that could likely be reduced via dictionary look ups or so.

I think we will have a Beta released next week, share with Mia, and allow her to click through the app and actually generate tickets for us beyond Create API/Frontend. 

As you are off for the day, I will be merging this in, and we can make a plan in sprint planning on Monday.

PR #263, opened 2025-07-16
bcwat-258-yearly-dropdowns

idaeusice:
# Description
Added front end functionality to support the yearly dropdown on charts. 

## Types of changes

## Checklist


LiamF-FS (commented):
Going over this - all is looking great.

Initial requests are to ensure the tooltip on the chart fits within viewport - when nearing right of x-axis, it gets cut off/pushed off screen.

Secondly, I feel as though we order the years chronologically, not based upon the order in which they were selected.

These are very minor things and not blocking. The functionality is fantastic. I am just going over the code now!

idaeusice:
> This looks good - the only thing that maybe could cause confusion is if someone clicks year B first, it gets assigned blue. They then select year A, which comes before year B, and then year A gets the blue colour, and year B gets a new colour.
> 
> It could cause confusion - I get there is probably some mapping of colours to year index. Everything else looks good :D

Agree that this could cause some confusion - something to think about for later iterations, however, as the other water tools appear to use this same mapping functionality. Good note.

PR #324, opened 2025-07-21
Update DAGs and the template that they use

### LINKED ISSUE #319: Adjust Template types for dags
# Purpose

Currently, most dags are assigned to use the `medium_task_template` or the `heavy_task_template`. This can be definitely changed. 

The purpose of the task is to run each dag, look at their memory usage, and change the template to be the appropriate one for it's memory usage.

Kashike-Umemura:
# Description

Previously the DAGs all used the `medium-task-template` which has 2Gb of memory allocated to it. This is overkill for most tasks.

The way that each dag was tested for their memory usage is the following:
- Make sure task does not OOM on OKD
- After task finish running, check for the maximum memory usage
- Assign appropriate task with ~200mb overhead just in case.

To optimize the memory allocation, a `tiny-task-template` has been created, which is `500Mib` of memory allocated to it, and the `small-task-template` has been changed to `750Mib` of memory.

## Types of changes

## Further comments
Closes #319


airflow/etl_pipelines/utils/functions.py
```diff
@@ -178,17 +178,17 @@ def update_station_status_id(db_conn = None):
...
                 UNION
                 (SELECT
                     station_id
                 FROM
                     bcwat_obs.climate_msp
                 WHERE
-                    datetimestamp >= (now() - '3 days'::INTERVAL)::DATE
+                    datestamp >= (now() - '3 days'::INTERVAL)::DATE
                 GROUP BY
-                    station_id) AS msp_stations
```
LiamF-FS, line 190: verifying this will not impact the DAG?

Kashike-Umemura: It did not. In fact with those in the query it failed, I think this will fix it

LiamF-FS (approved):
LGTM!

PR #333, opened 2025-07-22
Bcwat 321 remove licences from ne sw bc

### LINKED ISSUE #320: Make adjustment to fwa_funds so that the project is included in the table
# Purpose

There are slightly different allocation calculation logic in the watershed report. To accomodate this, the water sheds have to be assigned to the correct projects. And if there is an overlap, this can be delt with in the api.

The goal is to change the import a little bit so that the watershed_feature_id all has an region associated with it.

## Completion Criteria
This task is considered complete when the import script has been changed to have region_id with the wfi, and import has been ran.

### LINKED ISSUE #321: Remove licences from NE BC and SW BC
# Purpose

NE, adn SW BC does not have support for watershed licenesing functionality. So it should be removed from that.
Furthermore, the watershed report functionality should not exist in these regions either, so make a polygon that the click has to be within before it get's the report data.

## Completion Criteria
This task is considered complete when the licences get filter to the polygons that supports the watershed report function. And a mask of where the watershed report functionality is supported is created.

Kashike-Umemura:
# Description

Two issues are completed in this PR:
- Removing licences from North East BC and South/Coastal BC (NEWT, and Thompson River/Okanagan). This is because the area is not covered by watershed reports by any of the tools that are being consolidated. A map box layer has been made for this as well so that the mask covering the rest of the world can extend a bit to include the non-included parts of BC.
- Adding `region_id` value to the `fwa_fund` table. This is required because the calculation of the watershed reports value is slightly different depending on the region. The `region_id` value will allow us to easily determine which calculation method to use.

Other minor changes:
- Changed `quarterly_moe_hydrometric_historic_dag.py` template from medium to heavy since it was OOMing.

## Types of changes

## Checklist

## Further comments

Closes #321 
Closes #320


LiamF-FS (approved):
All of this looks great to me!

PR #336, opened 2025-07-22
Minor scraper fixes

Kashike-Umemura:
Minor changes to fix errors occuring in the scraper DAGs:
- Change min ratio to be a float value instead of dict.
- Asked Ben what these values should be, and he said that 0 is fine, since if the BC Gov want something more specific, they can change it themselves.
- Column that was being referenced was wrong, so fixed.


LiamF-FS (approved):
LGTM!

LiamF-FS (approved):
LGTM!

PR #339, opened 2025-07-22
Deploy e2e app onto openshift

LiamF-FS:
# Description

Changes initialized for openshift deployment.

Leaving PR Open so the deployment stays up. 

Wanting to ensure main/dev are at the same point to prevent merging issues in the future,


Kashike-Umemura (approved):
Looks great! Excited to have a dev env on openshift be up and running!

PR #341, opened 2025-07-23
Another scraper fix

Kashike-Umemura:
Minor fix to the ems dag to point to the correct directory to save the file to.
Also changed what the `licence_wls_map` a little bit. It actually had the `start_date` column but I messed up the original code so fixed it up.


LiamF-FS (approved):
Brilliant, all looks good to me!

Kashike-Umemura:
Removed `region_id` from the `bcwat_ws.fwa_fund` table because pre-calculation of this column will take waay too long

LiamF-FS:
> Removed `region_id` from the `bcwat_ws.fwa_fund` table because pre-calculation of this column will take waay too long

Makes sense to me!

PR #351, opened 2025-07-23
Return Terms of Monthly Mean Flow

LiamF-FS:
# Description

Update Response for streamflow monthly to return:

```
        "monthlyMeanFlow": {
            "years": monthly_mean_flow_year,
            "terms": monthly_mean_flow_term
        },
```

Both of these are a list of dictionaries, that include the yearly flow by month, and then the max/min/mean flow by month (terms)


client/src/components/MonthlyMeanFlowTable.vue
```diff
@@ -1,5 +1,6 @@
 <template>
     <div v-if="tableData">
+        {{ tableData }}
```
LiamF-FS, line 3: oop I will remove this

idaeusice (approved):
Looks great - I included the front end change to include terms in the monthly mean flow table.

PR #353, opened 2025-07-23
350 climate module changes july 21

MaxDouglas1991:
Change title of sidebar in climate page
Add paragraph to filters component

Station detail already includes temperature, precipitation and snow indicators
remove network and include temperature, precipitation and snow indicators in the station list

Make 10-90 colours brighter in chart legend to distinguish them
Change current colour to orange on the temperature chart
Remove current value from data after the current day
add 'daily mean' to year in tooltip

An issue was reported with the current values on the precipitation chart. The chart appears to always be getting data with values from decades ago instead of current values.

Prevent tooltip from doubling up on historical values
Make report tabs unselectable when station doesn't have data for them


client/src/components/MapFilters.vue
```diff
@@ -327,12 +328,29 @@
...
+                                dense
+                            >
+                                Precipitation
+                            </q-chip>
+                            <q-chip
+                                v-if="Object.keys(JSON.parse(item.properties.analysesObj)).includes('5')"
+                                dense
+                            >
+                                Snow Data
+                            </q-chip>
```
LiamF-FS, line 352: These .includes are a little off I think - I believe temperature is 6, 8.

Precipitation is 27.

Snow Data is 16, 19.

LiamF-FS: This is going off the data that goes into these charts - IE, the way the API breaks it down

client/src/components/climate/ClimateReport.vue
```diff
@@ -47,34 +47,39 @@
...
                     :class="viewPage === 'temperature' ? 'active' : ''"
+                    :disable="temperatureChartData.filter(entry => entry.currentMax !== null).length < 1"
                     @click="() => (viewPage = 'temperature')"
                 >
                     <div class="text-h6">Temperature</div>
                 </q-item>
                 <q-item
                     clickable
                     :class="viewPage === 'precipitation' ? 'active' : ''"
+                    :disable="precipitationChartData.filter(entry => entry.currentMax !== null).length < 1"
```
LiamF-FS, line 58: would you prefer if I passed back explicit keys declaring if the object has data for this module?

LiamF-FS: basically all of these disable checks

LiamF-FS:
be sure to merge dev back into this

client/src/components/climate/ClimateReport.vue
idaeusice: This is what we discussed yesterday, where we would add in flags for the different modules' availability. I think this makes sense to do.

MaxDouglas1991: I'm good with that

PR #370, opened 2025-07-25
Shade cells based on their hierarchy in row

MaxDouglas1991:
Shade values in mean monthly flow table by the values position relative to the min and max values. In the streamflow page the higher values get the deeper colours, while in the groundwater page the lower values get the deeper colours


LiamF-FS (commented):
This looks great!

LiamF-FS (approved):
Working as expected, I meant to approved with my other comment.

PR #375, opened 2025-07-28
first attempt at keeping db conns alive

NathanEne:
Attempting to keep the db connections alive. There is presently no ticket related to this work (sorry!)

### DIFF
--- a/backend/database.py
+++ b/backend/database.py
@@ -21,7 +21,7 @@ class Database:
         host = os.getenv("POSTGRES_HOST")
 
         try:
-            self.pool = ThreadedConnectionPool(minconn=1, maxconn=10, host = host, database = database, user = user, password = password, port = port)
+            self.pool = ThreadedConnectionPool(minconn=1, maxconn=5, host = host, database = database, user = user, password = password, port = port, keepalives=1, keepalives_idle=30, keepalives_interval=10, keepalives_count=5)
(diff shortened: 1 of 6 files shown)


LiamF-FS:
I shouldve made you a ticket - sorry about that!

backend/database.py
```diff
@@ -21,7 +21,7 @@ def __init__(self):
         host = os.getenv("POSTGRES_HOST")
 
         try:
-            self.pool = ThreadedConnectionPool(minconn=1, maxconn=10, host = host, database = database, user = user, password = password, port = port)
+            self.pool = ThreadedConnectionPool(minconn=1, maxconn=5, host = host, database = database, user = user, password = password, port = port, keepalives=1, keepalives_idle=30, keepalives_interval=10, keepalives_count=5)
```
LiamF-FS, line 24: why reduce max_conn? is that per pool? IE per worker?

NathanEne: yes, discussed a bit in mattermost. TLDR we do not expect that many ocncurrent users and this lets us give each connection more memory

LiamF-FS (commented):
Left some comments - my main concern is regarding is if the params we have entered are valid crunchy params or not.

LiamF-FS (approved):
Applied your yaml on OKD, all seems to have worked properly according to the `patronictl show-config` command!

Issue #364, opened 2025-07-25
Fix Mean Monthly Flow Calculation


LiamF-FS:
# Purpose

For both the streamflow/groundwater level mean monthly flow responses, ensure that the max/min/mean are properly calculated.

This will be within their respective files within the `utils` directory. 

The `min`/`max`/`mean` values should be the minimum of each month as a whole, instead of values within the month.

IE, from looking at KWT, these calculations are incorrect.

<img width="234" height="532" alt="Image" src="https://github.com/user-attachments/assets/c75bfc7a-3d55-4ab9-9ea2-dc119973e5a5" />

PR #380, opened 2025-07-28
Fixed up streamflow and groundwater means to return metrics based on average per month

AndersFarrFoundry:
# Description

Small change to the streamflow and groundwater metrics to ensure that the aggregate metrics (min, max and mean) are calculated based on the monthly mean, instead of the monthly overall data set. This makes these values clearer to users when interacting with streamflow and groundwater reports as these values will be generated from the means displayed in the table.

Open issue: https://github.com/bcgov/nr-bcwat/issues/364

## Types of changes

## Checklist

- [ ] I have added necessary documentation (if appropriate)


LiamF-FS (approved):
Excellent, simply smooth fix and the tests look great. Nice work!

Issue #366, opened 2025-07-25
Create Mean Annual Flow Calculation - Streamflow


LiamF-FS:
# Purpose

When requesting a single station on the streamflow module, we need to return the mean annual flow calculation.

<img width="346" height="319" alt="Image" src="https://github.com/user-attachments/assets/960e4304-57d6-4793-aa6d-cd0b0bcf8eb6" />

The data point will be calculated via one of 2 cases.

If there is less than 10 years of full data (IE a single month in a year has null data, it is NOT full), the mean annual flow will be the average of EVERY data point.

Otherwise, if there are at least 10 years of full data, the mean annual flow will be the averages of the average of each year.

This will be returned via the `@streamflow.route('/stations/<int:id>/report', methods=['GET'])` route.

PR #381, opened 2025-07-29
Mean annual streamflow calculation for streamflow report page

AndersFarrFoundry:
# Description

Calculate the mean annual streamflow (discharge) calculation for the streamflow report page using the following algorithm:

```
        If there are less than 10 years of full data (IE a single month in a year has null data, it is NOT full),
        the mean annual flow will be the average of EVERY data point.

        Otherwise, if there are at least 10 years of full data,
        the mean annual flow will be the averages of the average of each year.
```

This value is added to the stations/streamflow/{id}/report endpoint as part of its response.

I also added some empty data handling for one of the stations I found which didn't have any data for streamflow across the endpoint, and I snake_case'd a couple of variables in the streamflow endpoint.

Finally, I wrote the unit test test_generate_mean_annual_flow. I had to build some large cases to test the 10 full year behavior, but I do think that these tests effectively tests these cases.

https://github.com/bcgov/nr-bcwat/issues/366

## Types of changes

## Checklist


backend/routers/streamflow.py
```diff
@@ -52,17 +50,18 @@ def get_streamflow_station_report_by_id(id):
...
     raw_streamflow_station_metrics = app.db.get_streamflow_station_report_by_id(station_id=id)
     raw_streamflow_flow_metrics = app.db.get_streamflow_station_flow_metrics_by_id(station_id=id)
 
-    hasStationMetrics = raw_streamflow_station_metrics is not None
-    hasFlowMetrics = raw_streamflow_flow_metrics is not None
+    has_station_metrics = raw_streamflow_station_metrics is not None
+    has_flow_metrics = raw_streamflow_flow_metrics is not None
 
-    if not hasStationMetrics and not hasFlowMetrics:
+    if not has_station_metrics and not has_flow_metrics:
```
LiamF-FS, line 64: good call making this snake

backend/tests/unit/utils/test_streamflow_utils.py
```diff
@@ -1,5 +1,5 @@
-from datetime import datetime
-from utils.streamflow import generate_monthly_mean_flow_by_term
+from datetime import datetime, timedelta
+from utils.streamflow import generate_mean_annual_flow, generate_monthly_mean_flow_by_term
```
LiamF-FS, line 2: Very nit picky - I have been trying (and failing usually) to have different imports from the same file organized via:

```
from x.y import (
func_1,
func_2
)
```

backend/utils/streamflow.py
```diff
@@ -96,9 +96,10 @@ def generate_flow_duration_total_runoff(metrics: pl.LazyFrame) -> list[dict]:
         .collect()
     )
     min_year, max_year = year_bounds[0, "min_year"], year_bounds[0, "max_year"]
-
+    if(min_year is None and max_year is None):
+        # No data
+        return {}
```
LiamF-FS, line 101: This function is going to get removed when dev gets merged in, which will result in merge conflicts, just a heads up

backend/tests/unit/utils/test_streamflow_utils.py
AndersFarrFoundry: Definitely I can update with that format

backend/utils/streamflow.py
AndersFarrFoundry: I will merge dev in with my next push

backend/utils/streamflow.py
```diff
@@ -425,3 +428,98 @@ def generate_flow_metrics(flow_metrics) -> list[dict]:
...
+                ~pl.col("null_year")
+            )
+            .group_by(["year"])
+            .agg(
+                pl.col("value").mean().alias("mean_year")
+            )
+            .select(
+                pl.col("mean_year").mean()
+            ).collect()["mean_year"][0]
+        )
```
LiamF-FS, line 517: Can you explain quickly how this works with the null years? we join with null years to then filter out all of the years that have any null entry to ensure they are not taken into account?

And then we determine the mean of each year, and then the means of all the mean years? 

It makes sense - the only confusing piece reading this is joining with null years and then dropping null years

backend/utils/streamflow.py
```diff
@@ -425,3 +428,98 @@ def generate_flow_metrics(flow_metrics) -> list[dict]:
...
+    else:
+        return (
+            flow_metrics_lf
+            .select(
+                pl.col("value").mean()
+            )
+        ).collect()["value"][0]
```
LiamF-FS, line 524: Just validating - this does not take into account nulls correct?

client/src/components/streamflow/StreamflowReport.vue
```diff
@@ -41,7 +41,7 @@
                 </div>
                 <div v-if="'net' in props.activePoint" class="col">
                     <div class="text-h6">Mean Annual Discharge</div>
-                    <p>{{ props.activePoint.net }} m<sup>3</sup>/s</p>
+                    <p>{{ props.reportData?.meanAnnualFlow }} m<sup>3</sup>/s</p>
```
LiamF-FS, line 44: Maybe round this off to 2 decimals or so (`toFixed(2)`)

LiamF-FS (approved):
Left some comments - this looks brilliant. Nothing blocking, just clarification and comments! Excellent work!!! The tests are really valuable and well done, and offer clarity on expectations of function. A really good standard for the other util functions within the application.

backend/utils/streamflow.py
AndersFarrFoundry: Yes so the reason we join is that we have already calculated which years are NULL earlier in the function so that data is then merged with the current data and any data points which have their year value are then filtered by the below line. Essentially, I'm using the null_years that we determined above to figure out the full number of years as a filter, but I need them joined in in the first place to access that data.

Then yes, mean each year and then mean the means to get the overall as described in the methodology of the ticket

PR #383, opened 2025-07-29
Rewrite Flow Duration Tool

MaxDouglas1991:
Rewrite FDT using reductio and crossfilter to avoid making repeated API calls for filtered data


backend/routers/streamflow.py
```diff
@@ -59,7 +60,7 @@ def get_streamflow_station_report_by_id(id):
     raw_streamflow_station_metrics = app.db.get_streamflow_station_report_by_id(station_id=id)
     raw_streamflow_flow_metrics = app.db.get_streamflow_station_flow_metrics_by_id(station_id=id)
 
-    hasStationMetrics = raw_streamflow_station_metrics is not None
+    hasStationMetrics = raw_streamflow_flow_metrics is not None
```
LiamF-FS, line 63: this is incorrect, it should be reverted. I can do this

backend/routers/streamflow.py
```diff
@@ -123,6 +124,7 @@ def get_streamflow_station_report_by_id(id):
         "flowDuration":  computed_streamflow_station_metrics['flowDuration'],
```
LiamF-FS, line 124: flowDuration and all associated functions/routes can be removed if no API is called to calculate this stuff.

client/src/utils/api.js
```diff
@@ -53,6 +53,10 @@ export const getStreamflowReportDataById = async (id) => {
     return await requestWithErrorCatch(`${env.VITE_BASE_API_URL}/streamflow/stations/${id}/report`, 'report');
 }
 
+export const getFlowDurationByIdAndDateRange = async (id, startYear, endYear, month) => {
+    return await requestWithErrorCatch(`${env.VITE_BASE_API_URL}/streamflow/stations/${id}/report/flow-duration?start-year=${startYear}&end-year=${endYear}&month=${month}`, 'report');
+}
+
```
LiamF-FS, line 59: I do not think this gets used anymore

LiamF-FS (approved):
The flow duration stuff looks great. 

It is incredibly performant, seems to filter the data properly.

I left a few comments, please try to address them before merging in.

Great work!

backend/routers/streamflow.py
LiamF-FS: (I will remove these and push)

PR #387, opened 2025-07-29
Watershed Report - First Pass

LiamF-FS:
# Description

Initialize Watershed Report.

This is a first pass and required further work.

## Types of changes

## Further comments

We need to investigate behaviour for various regions, lakes, etc. 

We need to handle edge cases of missing data, and ensure the report displays the expected values for the region of interest.

To test this out, load the watersheds module, and click a point on the map. Some points will not have data - however, the points that do have data, you should see a watershed after clicking on the map, and hitting view report should generate a full report for that watershed.

### DIFF
--- a/backend/utils/surface_water.py
+++ b/backend/utils/surface_water.py
@@ -1,7 +1,24 @@
 import polars as pl
 
 def generate_chemistry(metrics: pl.LazyFrame) -> list[dict]:
-    return (
+    unique_params = (
+        metrics.
+        select(
+            pl.col("parameter_id")
+        )
+        .collect()
+        .n_unique()
+    )
+
+    sample_dates = (
+        metrics.
+        select(
+            pl.count("datetimestamp")
+        )
+        .collect()["datetimestamp"][0]
+    )
+
+    return ((
         metrics
         .with_columns(
             paramId=pl.col("parameter_id"),
@@ -16,7 +33,7 @@ def generate_chemistry(metrics: pl.LazyFrame) -> list[dict]:
             pl.struct(["d", "v"]).alias("data")
         ])
         .select("paramId", "units", "title", "data")
-    ).collect().to_dicts()
+    ).collect().to_dicts(), unique_params, sample_dates)
 
 def generate_surface_water_station_metrics(metrics: list[dict]) -> list[dict]:
 
@@ -36,6 +53,6 @@ def generate_surface_water_station_metrics(metrics: list[dict]) -> list[dict]:
         }
     )
 
-    chemistry = generate_chemistry(raw_metrics_lf)
+    (chemistry, unique_params, sample_dates)  = generate_chemistry(raw_metrics_lf)
 
-    return chemistry
+    return (chemistry, unique_params, sample_dates)
--- a/client/src/utils/api.js
+++ b/client/src/utils/api.js
@@ -123,9 +123,18 @@ export const getSurfaceWaterStations = async () => {
     return await requestWithErrorCatch(`${env.VITE_BASE_API_URL}/surface-water/stations`);
 }
 
+export const getSurfaceWaterStationStatistics = async (id) => {
+    return await requestWithErrorCatch(`${env.VITE_BASE_API_URL}/surface-water/stations/${id}/station-statistics`);
+}
+
 export const getGroundWaterStations = async () => {
     return await requestWithErrorCatch(`${env.VITE_BASE_API_URL}/groundwater/quality/stations`);
 }
+
+export const getGroundWaterStationStatistics = async (id) => {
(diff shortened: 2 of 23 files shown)


client/src/utils/api.js
```diff
@@ -24,18 +24,15 @@ export const getAllWatershedLicences = async () => {
     return await requestWithErrorCatch(`${env.VITE_BASE_API_URL}/watershed/licences`);
 }
 
+export const getWatershedBySearch = async (wls_id) => {
```
LiamF-FS, line 27: This should be titled getWatershedLicenceBySearch

LiamF-FS: another route should be mate for watershed search. I will update

backend/queries/watershed/get_licence_import_dates.py
```diff
@@ -0,0 +1,7 @@
+get_licence_import_dates_query = f"""
+    SELECT
+        dataset,
+        import_date
+    FROM
+        bcwat_lic.bc_data_import_date
```
Kashike-Umemura, line 6: You likely could filter this down since I put the hydat import date in this table as well. But it's also one row so no big deal imo

backend/queries/watershed/get_watershed_allocations_by_id.py
```diff
@@ -0,0 +1,6 @@
+get_watershed_allocations_by_id_query = """
+	SELECT
+    	*
+    FROM
+		bcwat_lic.get_allocs_per_wfi(%(watershed_feature_id)s)
```
Kashike-Umemura, line 5: I know that this probably returns a lot of columns, but I think it would be better if we select the specific columns and not `*` because then we would know if removing a column breaks some functionality in this query.
If you keep it as `*`, we wouldn't know if a column goes missing as easily.

backend/queries/watershed/get_watershed_annual_hydrology_by_id.py
```diff
@@ -0,0 +1,6 @@
+get_watershed_annual_hydrology_by_id_query = """
+    SELECT
+        *
+    FROM
+        bcwat_lic.get_annual_hydrology(%(watershed_feature_id)s);
```
Kashike-Umemura, line 5: Same as before

backend/queries/watershed/get_watershed_industry_allocations_by_id.py
```diff
@@ -0,0 +1,6 @@
+get_watershed_industry_allocations_by_id_query = """
+    SELECT
+        *
+    FROM
+        bcwat_lic.get_allocs_by_industry(%(watershed_feature_id)s);
```
Kashike-Umemura, line 5: Same here, but might be less relevant since everything gets aggregated in to a JSON obj

backend/queries/watershed/get_watershed_monthly_hydrology_by_id.py
```diff
@@ -0,0 +1,6 @@
+get_watershed_monthly_hydrology_by_id_query = """
+    SELECT
+        *
+    FROM
+        bcwat_lic.get_monthly_hydrology(%(watershed_feature_id)s, %(in_basin)s, %(region_id)s);
```
Kashike-Umemura, line 5: Ditto

Kashike-Umemura (approved):
Overall looks great! I left some minor comments about the queries that don't need to be changed. But if a data model change could happen in the future, I think it would help to eliminate ambiguity

backend/routers/watershed.py
```diff
@@ -96,7 +130,113 @@ def get_watershed_report_by_id(id):
             id (int): Watershed ID.
     """
 
-    watershed_metadata = app.db.get_watershed_report_by_id(watershed_feature_id=id)
-    # print(json.dumps(watershed_metadata, indent=2))
+    region = app.db.get_watershed_region_by_id(watershed_feature_id=id)
+    print(region)
```
AndersFarrFoundry, line 134: rm print

backend/utils/surface_water.py
```diff
@@ -1,4 +1,6 @@
 import polars as pl
+import json
+from datetime import datetime
```
AndersFarrFoundry, line 3: Are these imports now necessary? It looks like the change in this file is minimal. I've probably been removing a lot because my vs code removes unused imports on save.

AndersFarrFoundry (approved):
lgtm I don't know nearly as much as Kash about the SQL side of things! Successfully generated some reports though!

backend/utils/surface_water.py
LiamF-FS: probably not... I will take a look

backend/routers/watershed.py
LiamF-FS: cheers!

PR #393, opened 2025-07-30
Streamflow gwl csv download

LiamF-FS:
# Description

Initialize Download CSV on both Streamflow/Groundwater Level Modules.

Click on a station, and click download data in the left side bar.

This should download a CSV containing all time series data for that station. 

This is implemented for Climate/Streamflow/Groundwater Level - there is a bit of duplicate code in here (specifically the queries) however, they all use the same util function.

NOTE - router and util tests will be added for this later. Apologies, I just want to get this stuff in and then start cleaning up the API.

## Types of changes


backend/queries/groundwater/get_groundwater_level_station_csv_by_id.py
```diff
@@ -0,0 +1,18 @@
...
+  FROM
+    bcwat_obs.station_observation so
+  JOIN
+		bcwat_obs.variable v
+  USING
+		(variable_id)
+  WHERE
+    so.station_id = %(station_id)s
+  AND
+    so.variable_id IN (3)
```
AndersFarrFoundry, line 17: Total nitpick but you could just do = 3 here no?

backend/queries/streamflow/get_streamflow_station_csv_by_id.py
```diff
@@ -0,0 +1,18 @@
...
+  FROM
+    bcwat_obs.station_observation so
+  JOIN
+		bcwat_obs.variable v
+  USING
+		(variable_id)
+  WHERE
+    so.station_id = %(station_id)s
+  AND
+    so.variable_id IN (1, 2)
```
AndersFarrFoundry, line 17: I wonder if we should have a discussion about using ANY vs IN operators, they seem functionally the same but it seems like we interchange them. Not a big deal at all, but could be good to stick with one or another! Nothing to change here though!

AndersFarrFoundry (approved):
CSV looks good. I was wondering if we wanted to round off the numbers in the csvs, but honestly the users probably want the raw data to play with instead of it rounded so no issue. Great work!

backend/queries/groundwater/get_groundwater_level_station_csv_by_id.py
LiamF-FS: yes 100p I had just copied and pasted it good catch!

PR #396, opened 2025-07-30
Surface water groundwwater quality csv download

LiamF-FS:
# Description

Initialize CSV Downloads for Groundwater/Surface Water Modules.

This led to me adding a left join onto the metadata fetch for all CSVs, as these stations do not have entries within station_year columns.

There is a bit of formatting issue within the CSV, specifically introduced when a column includes commas/other special characters in its text. I am sure there is something I can do within polars to escape these such that they appear in a single row.

Again, testing coming soon!

## Types of changes


AndersFarrFoundry (approved):
LGTM as well, same as others!

PR #398, opened 2025-07-31
Reformat bar chart

MaxDouglas1991:
Reformat ReportChart and the data passed into it

All chart data is correct with dates lining up properly
Use current line / area with the appropriate values in the chart


LiamF-FS (approved):
I think we should get this into dev, and then make tickets from our progress meeting.

I included tickets that are probable to be created within my other comment, however,  I think getting the majority of our team's eyes on this would be beneficial to discuss expected behaviour.

There is an issue with the groundwater level chart that can be addressed in the future - we likely will want to update that to display in the same manner as KWT (monthly historical blocks)

PR #414, opened 2025-07-31
Search bar

MaxDouglas1991:
Update search bar to search for allocations and watersheds on Watershed page. 

Selecting an allocation selects is on the map, as do watersheds


LiamF-FS (approved):
Looks good to me! Search is working as expected for Licence number, as well as Watershed Feature Id. Great job!

PR #415, opened 2025-07-31
Bcwat 402 implement lake click

### LINKED ISSUE #402: Implement Lake click functionality
# Purpose

Lakes need to find the most downstream point that is still part of the lake, then get the upstream watershed from there. It currently does not do that, and only takes the upstream watershed located at the click.

OWT implements this so use that as a reference

## Completion Criteria

This issue is completed when clicking on a lake gets the upstream watershed of the most downstream portion of the lake.

Kashike-Umemura:
# Description

Found that the functionality that includes all the lake body in the watershed is caused by a different wfi being returned than the one that is being accessed. This will add a new column to the `bcwat_ws.fwa_fund` table that will allow us to select the right watershed.

## Types of changes

## Checklist

## Further comments
Closes #402


database_initialization/all_data_transfer.py
```diff
@@ -380,6 +380,7 @@ def import_from_s3():
             if table != "station_observation":
                 logger.debug("Truncating Destination Table before insert")
                 to_cur.execute(f"TRUNCATE TABLE {schema}.{table} CASCADE;")
+                to_conn.commit()
```
LiamF-FS, line 383: was this table not beign truncated before due to a missing commit?

Issue #403, opened 2025-07-31
Use Unique dates instead of each sample for sampling dates in bcwat streamflow and groundwater


AndersFarrFoundry:
Right now, we are using every single non-unique day for each sample to populate this data:

<img width="335" height="277" alt="Image" src="https://github.com/user-attachments/assets/240a2ce8-7ca9-4fd7-a567-99da1f368d60" />

This should have a "unique" flag on it, so the above would likely only have 1 day if all of the samples were taken on the same date.

PR #419, opened 2025-07-31
Changed logic to be a count distinct

AndersFarrFoundry:
# Description

Changed the logic of how we generate the sample dates for the groundwater and streamflow reports/pages to have distinct date counts instead of single days.

Updated unit tests as necessary.

https://github.com/bcgov/nr-bcwat/issues/403

## Types of changes

## Checklist

### DIFF
--- a/backend/queries/shared/get_water_quality_station_statistics.py
+++ b/backend/queries/shared/get_water_quality_station_statistics.py
@@ -1,9 +1,29 @@
 get_water_quality_station_statistics_query = """
+WITH merged AS (
+  SELECT
+    wqh.datetimestamp,
+    wqh.parameter_id
+  FROM
+    bcwat_obs.water_quality_hourly wqh
+  WHERE
+    wqh.station_id = %(station_id)s
+), count_dates AS(
+  SELECT
+    COUNT(*) as sample_dates
+  FROM
+    (SELECT DISTINCT datetimestamp FROM merged)
+  AS temp
+), count_params AS (
+  SELECT
+    COUNT(*) as unique_params
+  FROM
+    (SELECT DISTINCT parameter_id FROM merged)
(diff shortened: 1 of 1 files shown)


LiamF-FS:
This causes the Query to take a much longer time - but I am not sure if there is anything we can do about it. Unless we maybe group by date and count via polars in the API? I think it is the query that causes it to take so much time:

image

Implemented some logging in the API:

```
get_water_quality_station_statistics took 7.0044 seconds for station ID 43723
127.0.0.1 - - [31/Jul/2025 17:41:53] "GET /surface-water/stations/43723/station-statistics HTTP/1.1" 200 -
```

However, running the query in the DB itself takes less than half a second. So i am not sure what is causing this.

LiamF-FS:
The code also looks good! I am just a bit concerned about performance that I have encountered for the `ECCC - National Long-term Water Quality Monitoring Data` Network Surface Water Stations, as they have the majority of the points. But, again, running query in DB directly, performance is not inhibited.

LiamF-FS:
```
get_water_quality_station_statistics took 0.1948 seconds for station ID 43723
127.0.0.1 - - [31/Jul/2025 17:45:17] "GET /surface-water/stations/43723/station-statistics HTTP/1.1" 200 -
```

On hard refresh, the same request takes almost no time at all. Maybe its just our cluster being slow.

AndersFarrFoundry:
It was relatively performant for me yeah, I'm not really sure it may just be our cluster. I think we may have to see how it is on openshift?

AndersFarrFoundry:
I am going to refactor the query: https://stackoverflow.com/questions/11250253/postgresql-countdistinct-very-slow to try and improve performance! Will likely clutter it up a bit though!

PR #422, opened 2025-08-01
Fix pool

### LINKED ISSUE #421: Swap database connection pool to use que
# Purpose

presently when the pool is exhausted, add a que to prevent this!

NathanEne:
# Description

Swapped database pool to use SQLAlchemy for connection. This igives us access to some nice features such as connection recycling and checking the connection before using it. I will need some help testing this as I have run all the routes on both screens (dev vs this) and verified they line up but I dont totally know what I am looking for and how many of these errors are expected

## Types of changes

## Checklist

## Further comments

Closes #421


backend/database.py
```diff
@@ -15,75 +16,71 @@ def __init__(self):
         logger.info("Connecting to PostgreSQL Database...")
         port = os.getenv("POSTGRES_PORT")
         user = os.getenv("POSTGRES_USER")
-        password = os.getenv("POSTGRES_PASSWORD")
+        self.password = os.getenv("POSTGRES_PASSWORD")
         database = os.getenv("POSTGRES_DB")
         host = os.getenv("POSTGRES_HOST")
-
+        uri = f'postgresql+psycopg2://{user}:{self.password}@{host}:{port}/{database}'
```
LiamF-FS, line 22: we could just pass the uri in as an environment var - this is something that gets created on crunchy db user creation

LiamF-FS (commented):
I did approve this, but lets just keep this in PR for a while as we investigate time complexity of more intense queries.

NathanEne:
Sounds good, we will need to transform any new queries before this can get merged aswell

mishraomp:
we have had issues with Prisma connection pool in the past leading to memory leaks when primary dies and replica takes over, we switched to use pgbouncer for this reason. please make sure you do chaos testing before making this approach final

LiamF-FS:
We have initialized PGBouncer - therefore, the changes made by this ticket are unneeded.

Issue #417, opened 2025-07-31
Prevent Sub Modules from Displaying when no data


LiamF-FS:
# Purpose

We are properly displaying on the station interface the subModules in which a station has data for, however, on the subModule page itself, due to the structure and autofill functionality of the historical/current data structures regardless of it a station has data corresponding to a specific variable, we no longer indicate when a station has no data for a submodule.

<img width="1793" height="948" alt="Image" src="https://github.com/user-attachments/assets/c5e27025-bd05-4e5d-84d5-0de0b1d1dc82" />

<img width="1793" height="948" alt="Image" src="https://github.com/user-attachments/assets/62d08428-3c70-4b6c-b770-a6d84c89b780" />

In this case, Snow depth/SWE/Manual Snow Survey should all be disabled.

We can either do this through including a boolean flag that indicates the inclusion of a submodule for a certain station, or we can exit the historical/current loop if no the processed lazy frame is empty. in this case, the frontend could then just check the length of the array?

PR #424, opened 2025-08-01
Nr bcwat 417 report data filter out when data is not available

AndersFarrFoundry:
# Description

Filter out when data is not available for different aspects of reports (submodules). Use returned api flags where applicable, but mostly use a check on the data size of the returned data in a computed to set the disabled flag on the submodule click.

Also, don't render the submodule at all if there is no data available. 

Also, I made some small updates to the water quality (surface + ground) reports, to hide the chart when there is only a single data point, and no longer smooth the chart. 

I did some unit test changes here but to be honest a lot of those tests were so out of date that the changes I made stacked on top of already not working tests. Likely will just have to fix in the unit tests tickets. 

https://github.com/bcgov/nr-bcwat/issues/417

## Types of changes

## Checklist


backend/utils/shared.py
```diff
@@ -19,6 +19,8 @@ def generate_current_time_series(processed_metrics: pl.LazyFrame) -> list[dict]:
             eager=True
         ).alias("d")
     )
+    if(processed_metrics.collect().is_empty()):
+        return []
```
LiamF-FS, line 23: Excellent, super simple.

PR #425, opened 2025-08-01
Initialize Simple Cache for Station/Licenses Fetch

LiamF-FS:
# Description

Initialize Incredibly Simple Cache in API.

Only using this for big fetches - IE, initial data required on page load.

All individual station data will be fetched repeatedly, if you click on the same station repeatedly. We can change this.

You can validate this works as expected via opening up the browser, clicking on a page, clicking off, and going back to it. You should only see a single API request made, however, the data should be present on the FE.

## Types of changes


NathanEne (approved):
LGTM, I also tinkered with it and seemed to behave as expected. Definitely never going to be a harmful thing to have if people are switching tabs. If we move towards more long term caching we should keep in mind that the db does get updated with new stations occasionally.

LiamF-FS:
> LGTM, I also tinkered with it and seemed to behave as expected. Definitely never going to be a harmful thing to have if people are switching tabs. If we move towards more long term caching we should keep in mind that the db does get updated with new stations occasionally.

Yes I think an API cache would be good. We could update it pretty frequently, depending on when we get new stations. I think we should always be fetching the stations latest data in the API, but could move towards caching for specific ids for specific modules in the frontend.

PR #432, opened 2025-08-05
 Find closest watershed not necessarily the one intersecting so one is always found

AndersFarrFoundry:
# Description

For the get watershed by lat long endpoint, we were often getting a "watershed not found" because the stream we snapped our selected point to was not inside one of our watersheds. I updated the logic of this query to get all watersheds close to that point and then take the one with the smallest distance. Mostly this is about the feel of selecting a watershed on the front end, so the best test would be to click around and compare behavior between this tool and others (I mostly used OWT for comparison)

## Types of changes

Bug fix (non-breaking change which fixes an issue

## Checklist


NathanEne (approved):
LGTM

PR #433, opened 2025-08-05
408 fix watershed page lag on scroll

MaxDouglas1991:
Prevent each page from fetching all stations/allocations within the map view until the zoom value exceeds 9.
Change the message in filters sidebar to reflect whether the filter is being applied or not


AndersFarrFoundry (commented):
This seems to be working! I notice the streamflow page load is quite slow though, do you think we could implement a loading bar when the streamflow page is loading. I will send a message with what I mean more.

AndersFarrFoundry (approved):
Thank you for adding the loader!

PR #434, opened 2025-08-05
429 hydrology bar chart   incorrect display

MaxDouglas1991:
Update the values in the monthly hydrology chart to match existing charts


AndersFarrFoundry (approved):
LGTM. This looks really good. I think that the water allocations chart is different between this app and OWT but I think that is from the data side and I've created a ticket to look into it. Great job!

PR #442, opened 2025-08-05
Nr bcwat 438 data tab only

AndersFarrFoundry:
# Description

Added watchers to StreamflowReport.vue and ClimateReport.vue that will update the viewPage to only show a page with data available. Some other small fixes, fixed a console.error for climate reports without snow depth but with temperature.

To test, please check out different streamflow and climate reports with different data availability and see that a chart with data available is opened when the page is.

## Types of changes

## Checklist


MaxDouglas1991 (approved):
LGTM

PR #446, opened 2025-08-06
Reconfigure airflow secrets

LiamF-FS:
# Description

I may have branched off of the PGBouncer Test Branch, hence the large amount of changes.

I built this manually and have validated that the Dev Airflow suite is behaving properly. You can validate this by looking at the environment variables the scheduler/webserver/triggerer are initialized with.


NathanEne (approved):
LGTM

Issue #409, opened 2025-07-31
Watershed Report - Regional Fixes/Lakes


LiamF-FS:
# Purpose

- Region handling 
  - Certain regions omit candidates, etc
  - Certain regions include sections the others do not contain
  - Method section populated different per region

We must initialize handling within the API/FrontEnd for handling the various regional cases properly, coupled with handling the Lake case properly.

 
## Considerations

Requires FrontEnd/API Collaboration

The API should return flags indicating which region the report is for, coupled with flags indicating which sections are present for the report.

PR #450, opened 2025-08-06
Handle watershed report regional cases

AndersFarrFoundry:
# Description

Built off of Liam's work to return region ID and have different data calculated for hydrologic variability for different regions. I went over the reports for NWWT, OWT, KWT and Cariboo and it looks like we're producing all of the data in those reports.

The largest addition I had to make was creating the future hydrologic variability chart that was based off of the existing chart but had enough differences that it made sense to make into a separate chart and table. First time working with d3 so please let me know if I did anything crazy.

I also created separate objects for each of the methodology sections so we can populate with all of the station's data if and when we find it.

DB has been down today so I'm gonna hold off having this up until I can fully test with the DB.

The only thing that was not completed in the ticket was the "Lakes" handling, but Kash is still working on the DB changes for that and it should be simple to add that info in later for him.

https://github.com/bcgov/nr-bcwat/issues/409

## Types of changes

New feature (non-breaking change which adds functionality) -->

## Checklist


Kashike-Umemura:
the `fix lakes reports` commit and `change batch size` commit each has a two line changes. One to the `get_watershed_by_lat_lng.py` file, and the other to the `ems_archive_update.py` file.

The former change will ensure that all lakes have watershed report data, and the latter is an change to the number of rows retrieved at a time to make sure that the scraper is not killed.

AndersFarrFoundry:
Thanks for the update Kash!

backend/routers/watershed.py
```diff
@@ -153,115 +153,149 @@ def get_watershed_report_by_id(id):
...
+          "busStopNames": [bus_stop['name'] for bus_stop in bus_stops],
+          "ppt_mon_hist": watershed_metadata.get("watershed_metadata", {}).get("ppt_monthly_hist", []),
+          "ppt_mon_fut_max": watershed_metadata.get("watershed_metadata", {}).get("ppt_monthly_future_max", []),
+          "ppt_mon_fut_min": watershed_metadata.get("watershed_metadata", {}).get("ppt_monthly_future_min", []),
+          "tave_mon_hist": watershed_metadata.get("watershed_metadata", {}).get("tave_monthly_hist", []),
+          "tave_mon_fut_max": watershed_metadata.get("watershed_metadata", {}).get("tave_monthly_future_max", []),
+          "tave_mon_fut_min": watershed_metadata.get("watershed_metadata", {}).get("tave_monthly_future_min", []),
+          "pas_mon_hist": watershed_metadata.get("watershed_metadata", {}).get("pas_monthly_hist", []),
+          "pas_mon_fut_max": watershed_metadata.get("watershed_metadata", {}).get("pas_monthly_future_max", []),
+          "pas_mon_fut_min": watershed_metadata.get("watershed_metadata", {}).get("pas_monthly_future_min", []),
```
LiamF-FS, line 203: I set this up - but I think we should have better handling for when this data "watershed_metadata" is None (was testing some lake handling and got bad data)

```
Weedon Lake 
WFI: 7612875
Eaglet Lake
WFI: 7689625
```

I think we should just return a 400 or something if we detect watershed_metadata is empty?

@Kashike-Umemura do we know if the ppt_mon_hist etc is regional? or should ALL watersheds have these data points?

AndersFarrFoundry:
Added some handling for edge cases and a flyway migration to update some areas that did not exist.

Issue #400, opened 2025-07-31
Update the mapbox mask to use the mask in the BC Gov Mapbox account


Kashike-Umemura:
# Purpose

1. Currently the BC mask polygon that BCWAT is using is served from the Foundry Spatial Mapbox account.
Since a Mapbox account for this project has been made, we need to switch over to their layer. You can get the layer information by looking the `Styles` section on their [website](https://console.mapbox.com/studio/).

2. Furthermore, the watershed reports will not be covered in all of BC, so a different Mapbox mask needs to be used for that page ONLY. The style for that page is named `bcwat_ws_report_mask`.

3. The access tokens also needs to be changed to their access token

All login information is in BitWarden with user name `bcwatertool`

## Completion Criteria

- [ ] 1. is complete
- [ ] 2. is complete
- [ ] 3. is complete

PR #451, opened 2025-08-07
swapping map styles to have the filter polygon

NathanEne:
# Description
Swapped what mapbox account we are using and what mapbox style the watershed page so it is limited only to the region we support. TO TEST: you will need to swap your env to contain the mapbox token in bitwarden mentioned in the tikcet: #400 please also review the styles in mapbox to confirm all 3 requirements of #400 are complete

## Types of changes

## Checklist

## Further comments
Closes: #400

### DIFF
--- a/client/src/components/climate/ClimatePage.vue
+++ b/client/src/components/climate/ClimatePage.vue
@@ -34,7 +34,7 @@
                     @select-point="(point) => activePoint = point.properties"
                 />
                 <Map
-                    :current-section="'climate'"
+                    current-section="climate"
                     @loaded="(map) => loadPoints(map)"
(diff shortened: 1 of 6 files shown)


client/src/components/climate/ClimatePage.vue
```diff
@@ -34,6 +34,7 @@
                     @select-point="(point) => activePoint = point.properties"
                 />
                 <Map
+                    :current-section="'climate'"
```
MaxDouglas1991, line 37: These can be simplified as `current-secition="climate"`

MaxDouglas1991 (approved):
LGTM

PR #458, opened 2025-08-07
386 pdf download of watershed report

MaxDouglas1991:
Allow the user to generate a pdf from the watershed report


LiamF-FS (approved):
Looks good to me, PDF downloads and displays all included data.

PR #464, opened 2025-08-08
Initializing Prod DB with replicas, defaulting pgbackrest off (for now)

LiamF-FS:
# Description

Simple Init of Prod Database, created via command within README. 

I will be automating this process in the future.

Furthermore, I will turn pgbackrest on in the future - I would like to experiment with it OnPrem before backing up to s3 from openshift

image


NathanEne (approved):
LGTM, can see this up and running and data import has succeeded

PR #467, opened 2025-08-08
adding flyway migration that updates `in_study_area` flag

Kashike-Umemura:
Noticed that some areas in Omineca didn't generate a watershed report on click. This was because the `in_study_area` flag wasn't properly set for some of the watersheds in the `fwa_funds` table.
This migration will fix those values.


LiamF-FS (approved):
Looks great!

PR #470, opened 2025-08-09
Continue automated airflow secret creation

LiamF-FS:
# Description

Create Airflow Sendgrid Connection via GHA, validate email sends on Failure!

(I had hardset a divide by 0 and deleted our sendgrid default connection)

This will be what we do on Openshift to build airflow out of the box via a single github action.

My only concern is having to do with the PVC creation for airflow data, which should be simple to just do manually.

NOTE - until this goes in, any PR that goes in before it would break airflow scrapers (if they fail, no email sneds)


Kashike-Umemura (approved):
LGTM! Thank you

PR #481, opened 2025-08-11
add row count log, and decreased batch count

Kashike-Umemura:
Added a little bit more logs to know what's happening and decreased the batch count


LiamF-FS (approved):
Good inclusion, LGTM!

PR #488, opened 2025-08-11
Openshift airflow cicd

LiamF-FS:
# Description

Created Github Secrets for Airflow Deployments onto Openshift

Mirror the `deployer.yml` github action

The workflow will be to create a PR, run the unit tests against airflow code base, and then on Merge in all of the environments will update in order:
- Dev -> Test -> Prod


NathanEne (approved):
LGTM, Lets get this yml in and keep an eye on it to see how it does. Not related to this PR specifically but I have noticed that both yml and yaml are used in this project, is this something we should unify or is it a non concern?

AndersFarrFoundry:
> LGTM, Lets get this yml in and keep an eye on it to see how it does. Not related to this PR specifically but I have noticed that both yml and yaml are used in this project, is this something we should unify or is it a non concern?

I try to always use yaml, but I think they compile the same from the k8s side.

AndersFarrFoundry:
Could be that bcgov mostly uses yml and we mostly use yaml

PR #489, opened 2025-08-11
Change allocations display. Add watershed loader

MaxDouglas1991:
Add loader to watershed page when loading a watershed polygon

Change how existing allocations are displayed. They now extent down the chart as a means of identifying how much of the available water has been allocated


LiamF-FS (changes requested):
image

This must bottom out at 0, it cannot go into the negatives and impact the month display

Issue #270, opened 2025-07-17
Write Unit tests for `DataBcPipeline.py`


Kashike-Umemura:
# Purpose

Write pytests for `DataBcPipeline.py` class and for every method that is part of it.

## Considerations

A dummy class will have to be initialized, use the mock class that is in the `etl_pipelines/tests/utils/mock_DataBcPipeline_obj.py`
This will have all the methods that the `DataBcPipeline` class has. No need to test the abstract methods (the ones with `return None`) Since they will be tested in the other sub-classes.

## Completion Criteria

Write tests for:
- [ ] Initialization
- [ ] `download_data`
- [ ] `_load_data_into_tables`
- [ ] `get_whole_table`
- [ ] `update_import_date`
- [ ] `_check_for_new_units`
- [ ] `transform_bc_wls_wrl_wra_data`

Kashike-Umemura:
Handing over to Nathan

PR #509, opened 2025-08-12
Tests for DataBC ABC

NathanEne:
# Description

Wrote 19 tests for the DataBC ABC within the scrapers. Utilized alot of mocking for database interaction! Tried to test as well as I could but its tricky with so many database oriented functions. Added one test that requires internet connection (interested to hear peoples thoughts on if this is good or not) and utilized @pytest.mark.parametrize() to prevent having to copy paste cases a bunch.

Requesting review from Kash because he is the original author of this class and from Anders because he has written most of the tests in this repository presently.

image

Unfortunately the above warning is caused by one of our dependencies (possibly updating it would fix this)

## Types of changes

## Checklist

## Further comments

Closes: #270

### DIFF
--- a/airflow/etl_pipelines/tests/test_databcpipeline_class.py
+++ b/airflow/etl_pipelines/tests/test_databcpipeline_class.py
@@ -117,9 +117,7 @@ def test_mocked_download_data(get_data_mock):
                                'geometry': ['POINT (1 2)', 'POINT (2 1)']
                            }))
 
-import pytest
-import polars as pl
-from unittest.mock import MagicMock, patch
+
 
 @pytest.mark.parametrize("truncate", [
     (True),


airflow/etl_pipelines/tests/test_databcpipeline_class.py
```diff
@@ -1,7 +1,483 @@
...
+            'SEAPLANE_ACCESS_IND':pl.String,
+            'TC_LID_CODE':pl.String,
+            'SEQUENCE_ID': pl.Int64,
+            'SE_ANNO_CAD_DATA': pl.String
+        }
+    }
+
+    pipeline = TestDataBcPipeline('Test bc-airport pipeline', {}, databc_layer_name = airport_layer_name, expected_dtype=expected_dtype)
+    assert pipeline.download_data() is None
+    assert pipeline._EtlPipeline__downloaded_data[airport_layer_name] is not None
```
AndersFarrFoundry, line 99: I think that the external download is definitely worth trying. I do wonder how this will run from within a github action. Obviously they have internet connection but they make have sort of networkpolicy type things that you can't just do a cURL or download from an unknown URL. We won't know until we get these running on push but will be interesting to see.

airflow/etl_pipelines/tests/test_databcpipeline_class.py
```diff
@@ -1,7 +1,483 @@
...
+
+    pipeline.db_conn = fake_conn
+
+    pipeline.update_import_date("test_dataset")
+
+    fake_conn.cursor.assert_called_once()
+    fake_cursor.execute.assert_called_once()
+    query_arg = fake_cursor.execute.call_args[0][0]
+    assert "UPDATE" in query_arg
+    assert "test_dataset" in query_arg
```
AndersFarrFoundry, line 225: this is cool you are actualy looking for the shape of the query itself here very nice

airflow/etl_pipelines/tests/test_databcpipeline_class.py
NathanEne: Heres hoping it will work but we will see.

PR #512, opened 2025-08-12
raising max connections

NathanEne:
Very simple change to raise the max number of connections our connection pool in pythin can have now that we have initialized pgbouncer to do the real pooling behind the scenes (maxconn in python now represents max length of the 'queue' per worker)

Closes: [#514](https://github.com/bcgov/nr-bcwat/issues/514)

### DIFF
--- a/backend/database.py
+++ b/backend/database.py
@@ -20,7 +20,7 @@ class Database:
         host = os.getenv("POSTGRES_HOST")
 
         try:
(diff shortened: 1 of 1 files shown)


AndersFarrFoundry (approved):
Yeah good call. I wonder if we will ever get the traffic to need a queue of 50! Maybe with the search endpoints and multiple people trying at the same time. LGTM

NathanEne:
yeah I really doubt it but we will see!

PR #513, opened 2025-08-12
Fix some bugs that were found during Demo practice

Kashike-Umemura:
# Description

There were some bugs that I noticed and so fixed it:
- The unit and the y-axis label for the `StreamFlow` module's `Stage` chart was wrong so fixed that
- The values for the monthly demand and supply table were not matching up so I fixed the SQL functions as well as did some rounding fixes
- Licence count was wrong and also using the wrong data so fixed that as well.

## Types of changes

## Checklist


LiamF-FS (approved):
All of these changes & Stored Procedure changes look great.

PR #515, opened 2025-08-12
Add ENV Vars to enabling dags by default

LiamF-FS:
# Description

Add enable dags by default to Openshift


Kashike-Umemura (approved):
Thank you~ Looks good! We tested this just now for OKD

PR #522, opened 2025-08-13
Display annual volume

MaxDouglas1991:
Add the annual volume to the Allocations By Industry table. The row with the highest value will take the full width of the row and all others will take a proportional amount.

Also tidied up a couple things in the report


LiamF-FS (approved):
All of these changes look good and seem to meet the ticket requirements, as well as resolving other issues found within the report. Nice work!

PR #533, opened 2025-08-14
Remove document list and add link to site with documents

MaxDouglas1991:
Remove Licence Details dropdown from Allocations table and instead include a link the the site where the documents can be found


Kashike-Umemura (approved):
LGTM! THank you!

PR #534, opened 2025-08-14
Hide legend when watershed doesn't have min/max

MaxDouglas1991:
Hide legend for Topography chart when there is no min/max values


Kashike-Umemura (approved):
LGTM! Thank you!

PR #539, opened 2025-08-14
Pdf fixes

LiamF-FS:
# Description

Resize S3 Charts in DOM before downloading the PDF
Resize Table Clone in onclone html2pdf component

To validate this works as expected, please generate reports in all regions and validate that they are good quality and contain all content. No charts should be cut off.


client/src/components/watershed/WatershedReport.vue
```diff
@@ -252,43 +252,208 @@ const scrollToSection = (id) => {
...
-                .then(pdf => {
-                    pdf.addPage();
-                })
-                .from(element)
-                .toContainer()
-                .toCanvas()
-                .toPdf();
+const resizeS3ForPDF = (elements) => {
+
+    let originalStates = []
```
MaxDouglas1991, line 258: This variable never gets reassigned so it should be const

MaxDouglas1991 (approved):
LGTM. I pushed some small changes

Kashike-Umemura (approved):
LGTM thank you for working on this!

PR #544, opened 2025-08-15
swapping bus stops

### LINKED ISSUE #519: Watershed Report Bus Stops doesn't include Final water body
#### Describe the bug

Watershed Report has bus stop style map which shows the downstream route to the ocean:

Image

But as you can see, this does not include Final water body, which, for this case is the Pacific Ocean. The expected result is something like: 

Image

Do note that different rivers may go to different Final Water Body. The logic is located in each tools API (KWT, NWWT, OWT, and Cariboo)

#### Version Number
All Env

#### To Reproduce

1. Generate Watershed Report

### LINKED ISSUE #540: Refactor bus stop logic
# Purpose
Refactor the logic for the bus stops into python to be more testable and maintainable and also to fix some bugs while you are at it!
## Completion Criteria

- [ ] Bus stop logic is written into python
- [ ] Tests are written for new function(s)

#### Additional context

related to

### LINKED ISSUE #543: Add Williston Lake into bus stops when necessary
# Purpose

Due to issues in the underlying FWA code system OWT needed to add a hard-coded value to append Williston Lake when water is flowing into Peace River, duplicate this behavior in BCWAT!

## Completion Criteria

- [x] Similar to OWT all flow into the williston Lake says so in bus stops
- [x] Tests are written confirming this is the case

### LINKED ISSUE #545: Temporarily remove test_wsc_hydrometric.py tests
# Purpose
Remove the 3 failing airflow tests in `test_wsc_hydrometric.py` until Kash has a chance to do them!

## Completion Criteria

- [x] Troublesome tests commented out
- [x] ALL airflow tests passing in GHA

NathanEne:
# Description

Swapping the existing bus stop generation to be in python to better match existing OWT bus stop generation and increase test-ability while fixing some bugs. To assist with review this is  the existing OWT logic: https://github.com/FoundrySpatial/wammi-owt/blob/main/api/watershed/%5Bwfi%5D/report/getBusStops.js. It is known that this 'Williston Lake' manual override creates issues in a small region south west of the mouth of Peace River, however this was present in OWT and stems from our source data with no clear way to fix it, a conversation with Ben is due (FWA: 200-948755-000000-000000-000000-000000-000000-000000-000000-000000-000000-000000-000000-000000-000000-000000-000000-000000-000000-000000-000000 encompasses all of Williston Lake and Peace River).

Anders if you could review the code and testing that ould be excellent.
Kash if you could verify the outcomes on the frontend are as you would expect that would be amazing.

## Types of changes

## Checklist

## Further comments
Closes #540 
Closes #543 
Closes #519 
Closes #545

### DIFF
--- a/backend/routers/watershed.py
+++ b/backend/routers/watershed.py
@@ -195,7 +195,12 @@ def get_watershed_report_by_id(id):
     fwa_string = app.db.get_watershed_fwa_by_id(watershed_feature_id=id)['fwa_watershed_code']
     fwa_string_list = build_fwa_list(fwa_string)
     bus_stop_names = app.db.get_watershed_bus_stops_by_ids(fwa_watershed_codes=fwa_string_list)
-    post_processed_bus_stop_names = post_process_bus_stops(bus_stop_names)
+    try:
+        post_processed_bus_stop_names = post_process_bus_stops(bus_stop_names)
+    except ValueError as e:
+        return {
+            "error": "No found FWA id for the selected watershed. Please try a different watershed."
+        }, 500
 
     if(not "watershed_metadata" in watershed_metadata.keys() or watershed_metadata["watershed_metadata"] is None):
         return response, 404
--- a/backend/tests/unit/utils/test_watershed_utils.py
+++ b/backend/tests/unit/utils/test_watershed_utils.py
@@ -656,6 +656,8 @@ def test_post_process_bus_stops():
     assert len(result) == 10
     assert result[0] == "First Named River"
     assert result[1] == "Second Named River"
+    assert result[2:9] == ["Unnamed Basin"] * 7
+    assert result[-1] == "Pacific Ocean"
 
     bus_stops = [{"fwa_watershed_code": "100-123456-123456-000000-000000", "name": f"River {i}"} for i in range(15)]
     result = post_process_bus_stops(bus_stops)


backend/tests/unit/mock_database.py
```diff
@@ -362,21 +362,42 @@ def get_watershed_report_by_id(self, **args):
...
+            return bus_stops
+        elif watershed_id == fwa_id_9191927['fwa_watershed_code']:
+            from fixtures.watershed.wfi_9191927_bus_stops import bus_stops
+            return bus_stops
+        elif watershed_id == fwa_id_10255303['fwa_watershed_code']:
+            from fixtures.watershed.wfi_10255303_bus_stops import bus_stops
+            return bus_stops
+        elif watershed_id == fwa_id_9196070['fwa_watershed_code']:
+            from fixtures.watershed.wfi_9196070_bus_stops import bus_stops
+            return bus_stops
```
NathanEne, line 399: Sadly despite the code around this using the case statement we cannot here due to not having literal values but variables. It seems this is a limitation of case statments in python: https://stackoverflow.com/questions/66159432/how-to-use-values-stored-in-variables-as-case-patterns

If we want to I can swap this to use literal values instead of importing the other queries results as I am but this feels more clean and displays the logical link between these queries to future maintainers.

PR #558, opened 2025-08-18
added some limits to pgbackrest job

AndersFarrFoundry:
# Description

Added some limits to the pgbackrest job so we wouldn't have a huge spike in memory as we saw before. We were able to manually run a incremental and full backup and both worked in reasonable amounts of time.


NathanEne (approved):
LGTM, seems to be working

PR #561, opened 2025-08-19
changes to the document

Kashike-Umemura:
Your changes have been applied to this PR


NathanEne (approved):
LGTM, thanks! I dont believe this addresses the branching concerns but if that is not an issue in your opinion send this in

Kashike-Umemura:
> LGTM, thanks! I dont believe this addresses the branching concerns but if that is not an issue in your opinion send this in

Do you mean for the `fwa_watershed_codes`? I changed it to say that it 
> Forumlates the tree structure of the streams
I think this is more accurate than before but still might not be fully accurate

Kashike-Umemura:
oh yeah I'll check that as well

PR #562, opened 2025-08-20
548 uat   filters for search barsearching in the water module   abby morgan

MaxDouglas1991:
Add an API route to search for place names rather than query the API
Add indicator for current licence vs active application on watershed page
Remove network filter from watershed page. We will need to consult with Ben to see about adding the replacement filter values

### DIFF
--- a/backend/documentation/openapi.json
+++ b/backend/documentation/openapi.json
@@ -3246,6 +3246,97 @@
         }
       }
     },
+    "/watershed/location/search": {
+      "get": {
+        "tags": [
+          "Watershed"
+        ],
+        "summary": "Look up a location by their name",
+        "parameters": [
+          {
+            "name": "location_name",
+            "in": "query",
+            "required": true,
+            "schema": {
+              "type": "string",
+              "example": "Golden"
+            },
+            "description": "Location Name"
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "Payload returning licences matching query term",
+            "content": {
+              "application/json": {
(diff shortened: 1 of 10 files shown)


AndersFarrFoundry:
Will hold off review until Kash is done the APi unit tests

backend/documentation/openapi.json
```diff
@@ -3246,6 +3246,97 @@
...
+                      "type": "array"
+                    }
+                  }
+                }
+              }
+            }
+          }
+        }
+      }
+    },
```
AndersFarrFoundry, line 3339: Thank you so much for this Kash!

client/src/components/watershed/Watershed.vue
```diff
@@ -223,26 +225,26 @@ const watershedFilters = ref({
...
+        //         key: "net",
+        //         matches: "ERAA",
+        //     },
+        //     {
+        //         value: true,
+        //         label: "Canada Energy Regulator",
+        //         key: "net",
+        //         matches: "Canada Energy Regulator",
+        //     },
+        // ]
```
AndersFarrFoundry, line 247: maybe add a TODO here to ensure we know where to update after chatting with Ben?

AndersFarrFoundry (approved):
Code LGTM. I want to see how this runs on openshift as the speed of the search is critical. I think we could look into unbouncing (i.e. only searching after a user has stopped typing) instead of on each keystroke making an api call. Max, I'd love to discuss this with you.

PR #564, opened 2025-08-20
finished tests for functions.py

### LINKED ISSUE #272: Write tests for `functions.py`
# Purpose

The `functions.py` has some functions that don't belong in the scraper classes. They should be tested but they won't be too complicated

## Considerations

All of these db connections will come from the the airflow scrapers, so just make a mock database connection. Also, the `setup_logging` function does not need to be tested.

## Completion Criteria

Write tests for:
- [x] `update_station_year_table`
- [x] `update_station_variable_table`
- [x] `update_station_status_id`

#### Additional context
These tests might be pretty pointless but at least we can test that the exceptions are returned correctly if they fail

Kashike-Umemura:
# Description

Finished writing Tests for `functions.py`. It uses a `MockDbConn` class that is a `MagicMock` of the connection and cursor. I made this because I didn't want to flood the code with

```Python
fake_conn = MagicMock()
fake_cursor = MagicMock()

fake_conn.cursor.return_value = fake_cursor
```

So it looks a bit cleaner.
I also tried to initialize the `MockDbConn` object in the `conftest.py` file and import that, but then the `reset_mock` method didn't work as expected and it kept on throwing me errors.

## Types of changes

## Checklist

## Further comments

Closes #272

### DIFF
--- a/airflow/etl_pipelines/tests/test_functions.py
+++ b/airflow/etl_pipelines/tests/test_functions.py
@@ -49,7 +49,7 @@ def test_update_station_status_id():
 
     fake_db.cursor().execute.side_effect = Exception
 
-    # Test that the right exception get's raised if the excute fails:
+    # Test that the right exception gets raised if the execute fails:
     with pytest.raises(RuntimeError, match=r"Failed to update station_status_id in the table bcwat_obs\.station.*"):
         update_station_status_id(db_conn=fake_db.conn())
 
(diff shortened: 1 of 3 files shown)


airflow/etl_pipelines/tests/test_functions.py
```diff
@@ -0,0 +1,160 @@
...
+def test_update_station_status_id():
+    fake_db = MockDbConn()
+
+    # Test that exception is raised with no connection
+    with pytest.raises(RuntimeError, match=r"No database connection provided. Please provide a database connection.*"):
+        update_station_status_id()
+
+    fake_db.cursor().execute.side_effect = Exception
+
+    # Test that the right exception get's raised if the excute fails:
```
AndersFarrFoundry, line 52:
```suggestion
    # Test that the right exception gets raised if the execute fails:
```

AndersFarrFoundry (approved):
LGTM

PR #582, opened 2025-08-25
Remove outline from existing allocations chart. Change chart maximum

MaxDouglas1991:
Remove the outlines from the MAD bars so that the existing allocations are easier to see
Reduce the maximum size of d3 charts to be 1% higher than the maximum rather than 100% higher


MaxDouglas1991:
> Looks great! I saw that you said that the max Y value is 1% greater than the max, but the code says it's 5% greater than the max, just wanted to make sure that that's what you meant.

Sorry I was exaggerating for effect

PR #588, opened 2025-08-26
Unit tests for `water_approval_points.py`

### LINKED ISSUE #297: `water_approval_points.py`

Kashike-Umemura:
# Description

Unit tests added for `water_approval_points.py`! 
There are 5 csv files that are static data for input and verification of output. This method was chosen since the API tests are storing json objects in files for the verification of output as well.
Reading from the CSV files, there are some transformations needed to get the correct format of the DataFrame as the output. I tried to keep this to a Minimum, else we'd be comparing two very modified DataFrames, which may cause tests to fail on the verification side rather than the scraper side.

## Types of changes

## Checklist

## Further comments

Closes #297


AndersFarrFoundry (approved):
LGTM. It seems like some of these are becoming tricky to test. Like because it's just taking the data source and doing transformations, there are only so many cases. Good job getting what you could in there

PR #593, opened 2025-08-27
Unit tests for `asp.py` with other changes

### LINKED ISSUE #308: Write tests for `asp.py` climate scraper
This issue is considered complete when unit tests have been written for ALL functions in the class. Please Look at the parent issue for further explanation

Kashike-Umemura:
# Description

Unit tests for `asp.py` with other changes:

### Added the package [`callee`](https://callee.readthedocs.io/en/latest/) 
Has the use of `Contains` which allows it to check that a substring was in the log message. Without this I would need to match the whole string, which would be a mess because I have some tracebacks that get logged as well.
All scraper tests that could benefit from it had this added to it.

### `water_licences_bcer.py` missing `download_data` test
The `water_licences_bcer.py` test was missing the test for the `download_data` test, implemented that. Pretty happy with how it ended up.

### Scraper Fixes...
Finding some bugs that were in the scrapers that didn't get caught. Catching them as I write tests for them 

### 

## Types of changes

## Checklist

## Further comments

Closes #308


airflow/etl_pipelines/tests/test_asp.py
```diff
@@ -1,13 +1,250 @@
...
+    }
+
+    ## Set fake_check_for_new_stations to raise exception
+    fake_check_for_new_stations.side_effect = Exception("error")
+
+    pipeline.transform_data()
+
+    # Only going to check that the correct logs were logged
+    fake_logger.info.assert_any_call(f"Transforming downloaded data for {pipeline.name}")
+    fake_logger.error.asset_any_call(Contains("Failed to check for new stations in the downloaded data. Continuing on without checking."))
```
AndersFarrFoundry, line 170: I was going to dig more about this but it actually looks like the recommended way. Interesting that there aren't more conditional operations you can do here.

AndersFarrFoundry (approved):
LGTM, tests are working!

airflow/etl_pipelines/tests/test_asp.py
Kashike-Umemura: Yeah, the other option was to make a custom function that checks if the part of an argument existed. But this already does that and also has more versatile checking options so I think it's the best way for now

PR #596, opened 2025-08-28
`climate_ec_update.py` unit tests

### LINKED ISSUE #303: Write tests for `climate_ec_update.py` quarterly scraper
This issue is considered complete when unit tests have been written for ALL functions in the class. Please Look at the parent issue for further explanation

Kashike-Umemura:
# Description

Addition of unit tests for `climate_ec_update.py`, and some better logging.

## Types of changes

## Checklist

## Further comments

Closes #303


airflow/etl_pipelines/tests/test_climate_ec_update.py
```diff
@@ -1,7 +1,186 @@
...
+    )
+    plt.assert_frame_equal(
+        pl.select(pipeline.start_date),
+        pl.select(pl.datetime(
+            year=pendulum.now("UTC").subtract(days=92).year,
+            month=pendulum.now("UTC").subtract(days=92).month,
+            day=pendulum.now("UTC").subtract(days=92).day,
+            hour=pendulum.now("UTC").subtract(days=92).hour,
+            second=pendulum.now("UTC").subtract(days=92).second,
+            time_zone=str(pendulum.now("UTC").subtract(days=92).tz)
```
AndersFarrFoundry, line 91: Could these tests not use assert_frame_equal and instead just get the datetimes? Just wondering?

AndersFarrFoundry: Sorry if I missed asking about this in an earlier one!

AndersFarrFoundry (approved):
LGTM, just one comment for clarification

airflow/etl_pipelines/tests/test_climate_ec_update.py
Kashike-Umemura: They are using `assert_frames_equal`
The `pl.select` statement needs to be there because when an pl.datetime object is initialized, it isn't a dataframe but an expression, so the `pl.select` changes them into DFs.

I don't know if this answers your question, lmk if you have anymore!

AndersFarrFoundry: I was more wondering if you could do select then get the actual datetime itself and verify that it matches now - 92 days or whatever. I.e render it all the way out into a python datetime and verify its shape.

Kashike-Umemura: Something like this I presume 
image
Definitely works! I'll switch over to this

PR #601, opened 2025-08-28
ZAP Security Fixes

AndersFarrFoundry:
# Description

A bunch of changes to deal with issues in the ZAP security scanner:

1. Added root return of healthy to API when hit so no error is passed through
2. Added redirect to root url on not found values in vue router
3. Added sitemap.xml and robots.txt using [Sitemap](https://www.npmjs.com/package/vite-plugin-sitemap). Needed to create a new buildEnv.js to get the root url correctly.
4. Added sri vite package to auto-generate integrity field on npm build
5. Changed nginx.conf to not response to PATCH or TRACE requests, and also add the COOP header to responses.

## Types of changes

## Checklist

### DIFF
--- a/client/nginx.conf
+++ b/client/nginx.conf
@@ -39,6 +39,7 @@ http {
     add_header X-Content-Type-Options "nosniff";
     add_header Referrer-Policy "no-referrer";
     add_header X-Permitted-Cross-Domain-Policies "none";
+    add_header Cross-Origin-Opener-Policy "same-origin";
     add_header Content-Security-Policy "
         default-src 'self' http://localhost:* https://*.fspatial;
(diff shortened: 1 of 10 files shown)


client/nginx.conf
```diff
@@ -57,6 +58,10 @@ http {
 
     client_max_body_size 8M;
 
+    if ($request_method ~ ^(PATCH|TRACE)$) {
```
cetooley, line 61: I wonder if we want to include a comment here explaining the ZAP issue this resolves?

cetooley (approved):
I dunno if we need it not but some comments to explain certain things might help future devs too, however, as far as I'm concerned this is gtg

:shipit:

client/nginx.conf
AndersFarrFoundry: Good idea, I'll throw some more comments in

PR #605, opened 2025-09-02
Tag airflow with latest tag as well

AndersFarrFoundry:
# Description

When the test merge is completed, tag all images including airflow with the "latest" tag.

## Checklist

### DIFF
--- a/.github/workflows/prod_deploy.yaml
+++ b/.github/workflows/prod_deploy.yaml
@@ -49,7 +49,7 @@ jobs:
       packages: write
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


LiamF-FS (approved):
So this tags latest with whatever version was built via the merge? If so, excellent. I think we may need to add a similiar block to the `prod-deploy` that ensures a `prod` tag is built for Airflow.

After this meeting, I think we will have a better idea regarding our versioning.

AndersFarrFoundry:
Good point I added airflow to the prod re-tagging as well.

PR #609, opened 2025-09-02
Add client url to docker

LiamF-FS:
# Description

First Pass in reworking versioning of our various images for different environments.

As we need to pass in build_args to dictate the CLIENT_URL per Dockerfile, we cannot use a single build for multiple deploys.

This is to ensure that our ZAP Scans succeed per environment, as these changes were found to be required to produce the sitemap.xml & robots.txt

This is a first pass, and I would be happy to build specific versions of our test/prod images. 

I set it up to use `latest` for all airflow deployments, as these require a single build for multiple deployments.

This is untested, but we could wage the success of this after creating a PR into `main`


.github/workflows/prod_deploy.yaml
```diff
@@ -2,32 +2,47 @@ name: Prod Deploy
 
 on:
   workflow_dispatch:
-    inputs:
-      tag:
-        description: "Image tag set to deploy; e.g. PR number or latest. Use latest to re-tag test into prod"
-        type: string
-        default: 'latest'
```
AndersFarrFoundry, line 9: Honestly it's nice we don't even need to input this anymore

.github/workflows/prod_deploy.yaml
```diff
@@ -2,32 +2,47 @@ name: Prod Deploy
...
+        package: [backend, client, migrations]
+    timeout-minutes: 10
+    steps:
+      - uses: bcgov/action-builder-ghcr@v3.0.1
+        with:
+          keep_versions: 50
+          package: ${{ matrix.package }}
+          tags: prod
+          build_args:
+            CLIENT_URL=https://${{ github.event.repository.name }}-prod-frontend.apps.silver.devops.gov.bc.ca
```
AndersFarrFoundry, line 28: add a TODO here maybe as the URL is going to change to bcwatertool.ca at some point?

AndersFarrFoundry (approved):
LGTM, we will have to see how a deployment using this will goes. Seems like the full incarnation of what I was working on with a lot of improvements. 

We could create a tag set (re-tagging the "dev" tag) with the pr number if we wanted to have versioning as well? LGTM tho for now

.github/workflows/prod_deploy.yaml
LiamF-FS: Yes good point

LiamF-FS:
> LGTM, we will have to see how a deployment using this will goes. Seems like the full incarnation of what I was working on with a lot of improvements.
> 
> We could create a tag set (re-tagging the "dev" tag) with the pr number if we wanted to have versioning as well? LGTM tho for now

That is a good idea - I will make a note of this for a future iteration on this process, because some sort of tagging with versions beyond environment is likely best.

PR #615, opened 2025-09-03
Ensure deploy after build

LiamF-FS:
# Description

Set Deploy to run afterr build.

This still did work eventually, however, the deployments were in crash loops until the test images actually existed.


.github/workflows/merge.yml
```diff
@@ -21,7 +21,6 @@ permissions: {}
 jobs:
   build-airflow:
     name: Build Matrix
-    # needs: [unit_tests]
```
LiamF-FS, line 24: Removed this because unit tests run on PR creation - can add back in if you think necessary however

PR #618, opened 2025-09-04
607 flyway migrations

LiamF-FS:
# Description

The purpose of this PR is to rework when migrations occur within our pipeline. 

Previously, it was identified that applying database migrations to the dev database on PR Open was a major oversight and could cause extensive issues. Since this point, the team has reached an agreement that the best place for database migrations to occur is on merge into main - as this ensures that everyones dev database is in the same state. 

The only issue is that any erroneous migrations that are applied immediately go into the test environment - which could be mitigated against by splitting our pipeline into 2 target branches (dev/main) that will handle the various deployments. 

This was verified to have worked on OKD, if there are any issues with the dev/test/prod pipelines they should be easy to mitigate.

image

## Further comments

This also means that any sort of API changes that are dependent on a database migration must be written AFTER the migration has been applied to the dev database. IE, it will become a multi merge approach for introducing changes based on datamodel changes.

- Data Model Change Merged
  - If changing where API fetches data from, do not delete old data source!! 
- API Updates written
- Data Model cleaned up
  - This is where the old data source would be removed.

### DIFF
--- a/.github/workflows/.airflow-deployer.yml
+++ b/.github/workflows/.airflow-deployer.yml
@@ -31,7 +31,16 @@ on:
         description: Paths used to trigger a deployment; e.g. ('./airflow)
         required: false
         type: string
-
+      db_user:
+        description: The database user
+        required: false
+        default: bcwat-api-read-only
+        type: string
+      db_name:
+        description: The database name
+        required: false
+        type: string
+        default: bcwat-dev
       debug:
         description: Debug mode
         default: false
@@ -49,7 +58,6 @@ on:
         default: 'values.dev.yaml'
         required: false
         type: string
-
     outputs:
       tag:
         description: 'Which tag was used for deployment?'
@@ -57,7 +65,6 @@ on:
       triggered:
         description: 'Has a deployment has been triggered?'
         value: ${{ jobs.deploy.outputs.triggered }}
-
     secrets:
       oc_namespace:
         description: OpenShift namespace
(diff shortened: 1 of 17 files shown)


Kashike-Umemura (approved):
I assume that documentation needs to be updated? Specifically the `knowledge_transfer.md` docs. I can update it if, but if you want to then that sounds good as well.

LiamF-FS:
> I assume that documentation needs to be updated? Specifically the `knowledge_transfer.md` docs. I can update it if, but if you want to then that sounds good as well.

I will update it. I will also update the `CONTRIBUTING.md` etc

LiamF-FS:
> Code itself LGTM. I think this is very much going to force strictness and care with the deployments as you say because we are immediately going to deploy into test, even before API changes are written potentially. I think this could be very risky if not done with care as sometimes changes do not occur on a one-to-one basis where it is simply a column being changed. There may be times where a user needs to create an entirely new table to handle their data changes (imagine a new composite added to a primary key or something and the data is cartesian producted along a new dimension). I think with or without this developers would have to be very careful, and external contributors may find having to create multiple migrations to slowly migrate with a new column/table, etc. tedious. However, I don't have a better solution in front of mind, this whole auto-deployment per PR business just makes the whole thing very tricky. Honestly LGTM with reservations that are probably in-addressable atm

Yeah its a tough problem. I think the next time we meet with them we need to ask if they are content with this approach, or if they would like to divulge from the single branch driving all deployments, and instead set up a staging branch of sorts with the entire goal being protecting test/prod from having its data model go into a bad state. Will ensure to ask the BCGov folks at the next opportunity. Fortunately, I do not think that any sort of changes we need to make regarding where things should occur are difficult, as we just changing target branches on push.

LiamF-FS:
> (imagine a new composite added to a primary key or something and the data is cartesian producted along a new dimension)

I can put them onto this branch

PR #620, opened 2025-09-04
Create and Seed each DB via GHA

LiamF-FS:
# Description

The purpose of this PR is to create Github Actions to handle the disaster recovery of each database on Openshift.

It should be noted that for these to work - there must be S3 backups present within the specified bucket. Otherwise, if required, I could make a separate action that created an empty database and then populated it via a backup of a different environment - However, this seems like a long walk for something that will likely never be used:
- Requires creating an empty database in namespace X
- Requires creating a backup from a designated namespace Y
- Requires restoring from said backup to empty database in namespace X

This would be a very long running action - we do have the majority of the code for this already written, as it would just be kicking off the backup/restore jobs we previously created.

I have not validated that this works on openshift, however this process would have worked on OKD if the collation version of the OKD/Openshift Postgres' aligned.

We will chat with the BCGov folks in the future, as I would love to validate at least once that this approach works on the Dev namespace.


.github/workflows/dev-deploy-seed-db.yml
```diff
@@ -0,0 +1,28 @@
...
+    deploys:
+        name: Deploys Latest
+        uses: ./.github/workflows/.db-deployer.yml
+        secrets:
+            oc_namespace: ${{ secrets.OC_NAMESPACE }}
+            oc_token: ${{ secrets.OC_TOKEN }}
+            s3_access_key: ${{ secrets.S3_ACCESS_KEY }}
+            s3_secret_key: ${{ secrets.S3_SECRET_KEY }}
+            s3_bucket_name: ${{ secrets.S3_BUCKET_NAME }}
+            s3_endpoint_url: ${{ secrets.S3_ENDPOINT_URL }}
```
AndersFarrFoundry, line 23: Are these secrets going to be the same for every environment... i.e. are we always going to be restoring from a prod backup if we are doing this? Just checking not a big deal either way.

.github/workflows/prod-deploy-seed-db.yml
```diff
@@ -0,0 +1,23 @@
...
+  group: ${{ github.workflow }}
+  cancel-in-progress: false
+
+permissions: {}
+
+jobs:
+    deploys:
+        name: Deploys Latest
+        uses: ./.github/workflows/.db-deployer.yml
+        secrets: inherit
```
AndersFarrFoundry, line 17: Why does this one need secrets: inherit while the dev one manually specifies? Is this meant to be called from another place ?

AndersFarrFoundry (approved):
I think this looks good, it would've been nice if we could've had a successful run but still it isn't that complicated of an action! I just had some minor clarifications!

Kashike-Umemura (approved):
Thank you!

.github/workflows/prod-deploy-seed-db.yml
AndersFarrFoundry: Sorry I typed something and totally forgot to hit send. That makes sense that it should overwrite with those values. It does seem like there are multiple ways to do this here and directly specifying prod in the name maybe would be simpler but honestly it is basically the exact same level of complexity either way. All makes sense ty for clarifying!

PR #621, opened 2025-09-04
Fix Issue with Values Formatting

LiamF-FS:
# Description

Helm Deploy was failing as all references to `{{ .Values... }}` blocks within a values file must be wrapped in Quotes.


AndersFarrFoundry (approved):
LGTM good catch!

PR #624, opened 2025-09-05
Move init block into webserver (retain logs to debug if migrations fail)

LiamF-FS:
# Description

Move the ExtraInitContainers block into the webserver block - this is just so logs will retain and it will be easier to debug as the webserver is long running, whereas the migrateDatabaseJob does not retain logs.


Kashike-Umemura (approved):
Thank you for finding this!

PR #627, opened 2025-09-05
Airflow Migrations needs same image as airflow

LiamF-FS:
# Description

Fix Images Built for Airflow Migrations


Kashike-Umemura (approved):
LGMT

PR #642, opened 2025-09-09
Lock Airflow Version - Just Incase

LiamF-FS:
# Description

Locking Version of Airflow in values just in case this would cause an incompatibility in the future.


AndersFarrFoundry (approved):
Good idea!

PR #646, opened 2025-09-10
flyway migration and query updates

### LINKED ISSUE #536: Investigate Stations with `station_status_id` = 5 and change them accordingly
#### Describe the bug

There are stations in the 5 Non-Watershed modules of the BC Water Tool. These should not show up on the frontend, and the filters for them should be removed.

A little bit of investigation has been done, and it has been found that some of the stations DO infact have data. These stations should be changed to `Historical`.

#### To Reproduce

1. Go look with your eyes

#### Expected behavior

There are a few things that should happend:
- [ ] There should be NO GRAY DOTS on ANY MAP
- [ ] The stations currently marked as `station_status_id = 5` must be investigated if they have data. If they do, change to historical
- [ ] API has been edited to filter `Not Available` stations out

Kashike-Umemura:
# Description

Made sure that the stations that have status' `Not Available` actually doesn't have any data. If they do have data, they have been changed to `Historical` status.
Furthermore, the queries that collect all the stations does not fetch any stations that have `Not Available` as their status.

## Types of changes

## Checklist

## Further comments

Closes #536

### DIFF
--- a/backend/queries/climate/get_climate_stations.py
+++ b/backend/queries/climate/get_climate_stations.py
@@ -79,4 +79,6 @@ get_climate_stations_query = """
       s.type_id IN (3, 6)
     AND
       s.prov_terr_state_loc = 'BC'
+    AND
+      ss.status_name != 'Not Available'
 """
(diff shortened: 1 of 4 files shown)


LiamF-FS (approved):
Looks good - good find and fix!

PR #703, opened 2025-11-05
Ok - this SHOULD FIX EVERYTHING

LiamF-FS:
# Description

This should properly fix dev/test/prod Airflow deployments. 

Permissions have been granted for the new users on dev/test. I just had forgotten the bcwat db user needed to be the bcwat_airflow_read_write user.


Kashike-Umemura (approved):
Thank you!

PR #705, opened 2025-11-05
fix security contexts, test/prod

LiamF-FS:
The Test/Prod security contexts had been overwritten, stalling the migration jobs.

This fixes that and will ensure the next helm deploy runs as expected.


Kashike-Umemura (approved):
LGTM

PR #707, opened 2025-11-06
bcwat-666-groundwater-filters

idaeusice:
# Description
Updated filtering functionality to work as expected. There were several cases where filtering was having either no effect, or filtering out all points. 

## Types of changes
Set up a dynamic check on the points to get unique filterable properties and list those as filtering options, as well as separate expressions to build the other types of filtering: year range, quantity, area. 

## Checklist


LiamF-FS:
> image Turning filters off & on doesn't restore the map points in streamflow gauges

I also found this, for clicking any of the options. Wipes out all the points.

idaeusice:
The filters should work as expected now. Because of some substantial restructuring of the various water portal pages, the tests will need some updates.

PR #713, opened 2025-11-07
bcwat-667-pdf-illegible-text

idaeusice:
# Description
Updated the pdf report to be more readable for tables. 
Updated font color on tables to use the brownish font found on previous water tools. 
Added WFI to watershed report. 

## Checklist

## Further comments
It's possible that this could be improved further. Some of these tables are putting so much content in too small of a space, which impacts readability. However, this change allows gets the content to be displayed without the columns being one or two characters wide.

### DIFF
--- a/client/src/components/watershed/report/MonthlyHydrologyTable.vue
+++ b/client/src/components/watershed/report/MonthlyHydrologyTable.vue
@@ -161,7 +161,7 @@ const tableCols = [
         td,
         th {
             color: $table-font-color;
-            font-family: 'BC Sans';
+            font-family: 'BC Sans', sans-serif;
             text-align: end;
 
             &:first-child {


client/src/components/watershed/WatershedReport.vue
```diff
@@ -529,7 +520,7 @@ const pdfDownload = async () => {
                         svg.style.width = state.styleWidth || '';
                         svg.style.height = state.styleHeight || '';
                     } catch (error) {
-                        console.error(state.elementId)
+                        console.log(state.elementId)
```
cetooley: what's the reason for switching to log here?

client/src/components/watershed/WatershedReport.vue
```diff
@@ -399,29 +400,19 @@ const resizeS3ForPDF = (elements) => {
...
 
     targets.forEach(({ sel, max }) => {
         clonedDoc.querySelectorAll(sel).forEach((table) => {
-            table.style.width = '100%';
-            table.style.maxWidth = `${max}px`;
-            table.style.tableLayout = 'fixed';
-            table.style.marginLeft = 'auto';
-            table.style.marginRight = 'auto';
-
+            table.style.width = `100%`;
```
cetooley: Something that might be interesting to try in the future is to scale an element based on the size of the container
i.e. get the size of the element as layed out, get the container width, divide and then use that value for scaling: using https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/transform-function/scale

PR #718, opened 2025-11-14
bcwat-548-filtering-update

idaeusice:
# Description
Update the filtering to exclude network. 

## Types of changes
Updated watershed filtering: removed network, added some handling for potentially missing values that were otherwise uncaught

## Checklist


Kashike-Umemura (changes requested):
image

I'm still seeing the `Network` section for the watershed licence filters.
Also could you change the `Historical` indicator to say `Current Licences`? I think Historical would imply expired or inactive, and we don't have expired licences in the app.

Ty!

PR #723, opened 2025-11-18
BCWAT - Bug Fixes and Maintenance

LiamF-FS:
- Fix PDF
- Increase precision of watershed selection on the map
- New feature allowing users to download the watershed polygon in geojson or shapefile format.
- Fixed issue where allocation amount wasn't showing up for some licences.
- Added water source name to licences on the sidebar
- Various other Bug Fixes and Maintenance tasks

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "bc-wat-app",
-    "version": "4.10.4",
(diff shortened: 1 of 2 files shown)


idaeusice (approved):
Things are looking good. I included a small fix around the map filtering, where year range was not resetting correctly. It appears to be working now.

PR #730, opened 2025-11-24
686 temperature chart

LiamF-FS:
Changed way in which temperature chart is rendered. Display Max/Min values for yearly dropdown, change current values to be lines instead of shaded area.

All tests updated, one added in API, no test added for FrontEnd - tests already exist, however I likely could add a new case in for the temperatureChart.

### DIFF
--- a/client/src/components/ReportChart.vue
+++ b/client/src/components/ReportChart.vue
@@ -640,9 +640,6 @@ const addMedianLine = (scale = scaleY.value) => {
         );
 };
 
-const ensureCurrentLinesOnTop = () => {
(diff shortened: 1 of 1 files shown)


backend/tests/unit/routers/test_climate_router.py
```diff
@@ -112,6 +112,7 @@ def test_get_climate_station_report_temperature_by_id_and_year(client):
     assert response.status_code == 200
 
     data = json.loads(response.data)
+    print(json.dumps(data, indent=2))
```
Kashike-Umemura: Print call :))

PR #732, opened 2025-11-25
Add Limits to Airflow Services/backend/FE on Openshift

LiamF-FS:
# Description

Add Limits to our Openshift Airflow/Api/FrontEnd Pods. 

These values were gathered by looking at the resource usage across upper each pod and providing more than what has ever been used, just to ensure this does not disrupt the health of our namespaces up to this point.

### DIFF
--- a/charts/openshift/app/templates/frontend/templates/deployment.yaml
+++ b/charts/openshift/app/templates/frontend/templates/deployment.yaml
@@ -66,7 +66,10 @@ spec:
           resources:
             requests:
               cpu: 30m
-              memory: 50Mi
+              memory: 150Mi
+            limits:
+              cpu: 30m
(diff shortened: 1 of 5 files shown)


charts/openshift/app/templates/frontend/templates/deployment.yaml
```diff
@@ -66,7 +66,10 @@ spec:
           resources:
             requests:
               cpu: 30m
-              memory: 50Mi
+              memory: 150Mi
```
AndersFarrFoundry: Does the front end need this much? 150 MB seems excessive for an nginx container, but also like not a huge deal

AndersFarrFoundry (approved):
LGTM, tiniest comment about front end stuff

PR #746, opened 2025-12-09
Update okd api env

LiamF-FS:
# Description

This should rebuild our deployments using the new secrets, and mean that we can decommission crunchy on Prem. New Secret Exists.

Remove Crunchy Outright, I can helm uninstall the DB. 

Hopefully I made the jdbc uri correct, it is always hard to tell if escaping characters is required.


Kashike-Umemura (approved):
LGTM, Thank you for the updates

PR #753, opened 2025-12-30
Hotfix msp issue

AndersFarrFoundry:
# Description

When we were getting data for the msp scraper, it would have column headers with whitespace:

`Snow Course Name, Number, Elev. metres, Date of Survey, Snow Depth cm, Water Equiv. mm,  Survey Code, Snow Line Elev. m, Density %, Survey Period`

When these are then loaded from the csv into a polars lf, the schema_overrides are applied based on the headers stripped of whitespace. However, since these are not the current headers the schema isn't applied and therefore the schema can not match and the scraper can fail!

## Types of changes

## Checklist


Kashike-Umemura (approved):
Changes look good and tested! LGTM! Thank you~

PR #755, opened 2026-01-06
Excluded ASP from getting type casted in Download

Kashike-Umemura:
# Description

There was a recent fix that went in that added a type cast to the download function. This will break the ASP scraper because there are minor transformations that needs to be done BEFORE the data verification step. I missed this in the PR because ASP is the only scraper that requires this transformation.

## Types of changes

## Checklist


LiamF-FS (approved):
LGTM!

PR #773, opened 2026-03-03
Upgrade airflow version

LiamF-FS:
# Description

Preliminary Changes to bump Airflow Version.

This includes changes to the DAGs themselves, namely the import structure and the task declaration. 

The bulk of the changes are centered on Config. We now require a JWT Secret, and the Webserver has been deprecated. This is why there are now references to the API_SERVER, as this replaced the existing webserver.

I created a JWT Token/API Secret key in the github secrets, which should trigger the secrets to properly get created on OKD. Once this has been validated on OKD, I will make the same changes to the openshift pod templates/charts.


NathanEne (approved):
LGTM

PR #840, opened 2026-03-25
792 configure ches email backend

LiamF-FS:
# Description

- Toggle Email functionality based upon if on OKD (Sendgrid) or Openshift (CHES)
- Update Default Args object to be conditional from ENV
- Remove Sendgrid Information from all Openshift Pod Templates
- Send Email via `on_failure_callback` workflow
- Populate CHES credentials via Github environment secrets/variables
  - This creates our openshift secrets

Created a `failure_on_purpose` DAG, and CHES_TO_ADDRESS to be my email address to validate working as expected across all environments.


Kashike-Umemura (approved):
Looks good! Thank you for implementing their emailing system. There are some places that I would like to add email notifications to, but I think that's out of scope for your assigned work so possibly in the future. Thank yoU!

PR #853, opened 2026-04-07
add enmods

NathanEne:
## Types of changes

## Checklist

## Further comments

Updating EMS data source to EnMoDS data source Both quarterly and daily scrapers as per the client request. A few filed are missing so they have been filled with generic data. Otherwise it was largely based off EMS


airflow/dags/quarterly_enmods_water_quality_dag.py
```diff
@@ -11,23 +11,23 @@
 PLATFORM = os.getenv('PLATFORM', 'no-platform-found')
 
 @dag(
-    dag_id="quarterly_ems_water_quality_dag",
+    dag_id="quarterly_enmods_water_quality_dag",
     # Cron for At 08:30 UTC (00:30 PST) on day-of-month 2 in every 3rd month.
```
Kashike-Umemura: Could you remove the PST comment from this. It's wrong now and I think we should keep only the UTC one

airflow/etl_pipelines/utils/constants.py
```diff
@@ -1007,22 +1007,55 @@
 }
 QUARTERLY_MOE_HYDRO_HIST_MIN_RATIO = 0
 
-QUARTERLY_EMS_NAME = "Quarterly EMS Archive Update"
 QUARTERLY_EMS_CURRENT_URL = {
```
Kashike-Umemura: Any reason that this one is still here?

migrations/sql/V1.0.9__update_water_quality_location_types.sql
```diff
@@ -0,0 +1,88 @@
...
+--     SELECT station_id FROM bcwat_obs.station WHERE network_id = 25
+-- );
+
+-- DELETE FROM bcwat_obs.station_water_quality_parameter
+-- WHERE station_id IN (
+--     SELECT station_id FROM bcwat_obs.station WHERE network_id = 25
+-- );
+
+-- DELETE FROM bcwat_obs.station where network_id = 25;
+-- -- I imagine some of the ids changed so we don't want zombie stations in the database (just need to make sure to run the ENMODS DAGs immediately after this migration to repopulate the stations table with the new station ids)
```
Kashike-Umemura: We could remove this since it won't be ran. But I don't mind if it's kept in either

Kashike-Umemura (approved):
Thank you for implementing this! Everything looks great and we kind of have to test on Dev to make sure that it works anyways.
I've left some small comments on things that might be removable, but they are not blocking the merge.

airflow/etl_pipelines/utils/constants.py
NathanEne: nope, removing

migrations/sql/V1.0.9__update_water_quality_location_types.sql
NathanEne: will just rm due to our research

PR #858, opened 2026-04-14
New PVC Specs, Update EnMoDs resources, update pod templates

LiamF-FS:
# Description

- Create new PVC (more resources, RWX)
- Update Pod Templates
- Add Resources to EnMoDs daily


NathanEne (approved):
LGTM, thanks for doing this

PR #859, opened 2026-04-27
RM Everything OKD Specific

LiamF-FS:
Remove everything OKD specific from the BCWAT Deployment Pipelines.

This basically just means removing any OKD config, github actions, or workflow specific stuff.

I have also removed all of the Github Secrets that are no longer used.


Kashike-Umemura (approved):
Thank you for removing the unused files!

PR #868, opened 2026-05-21
Remove Analysis/Notifications, bump action versions to latest

LiamF-FS:
The purpose of this task is to bump the versions of our reusable github actions to no longer use any versions that use node 20.

While at it, I bumped some of the vresions of the bcwat reusable actions to get them on later available, while ensuring that the args we currently use will remain compatible.


AndersFarrFoundry (approved):
LGTM!

PR #869, opened 2026-05-22
Address pip vulnerabilities

LiamF-FS:
# Description

Address Vulnerabilities across:

- airflow
- backend
- client
- database_initialization
- pdf_generator

I have also updated our Dockerfiles to use `psycopg2` instead of `psycopg2-binary` import.

You can validate the application runs as expected via:

```
docker compose build
docker compose up
```

There should be no issues with DB Initialization/Airflow - I have validated that the ParquetFile functionality of `pyarrow` (the largest version bump) should not be broken by the new version.

We will be able to validate proper functionality of the application/airflow within the dev/test namespace IF this PR gets merged in.


NathanEne (approved):
LGTM

idaeusice (approved):
Updated front end with recent fixes for new vulns. Front end and pdf gen appear to be working as expected.

PR #872, opened 2026-06-02
Fix Ariflow Dockerfile - Necessary for Psycopg2

LiamF-FS:
Fix Docker File - issue building psycopg2.

https://github.com/bcgov/nr-bcwat/actions/runs/26839526620/job/79162399344

I do not want to use psycopg2 binary.


NathanEne (approved):
LGTM

PR #873, opened 2026-06-17
Testing Create Airflow Public User @ Public Route

LiamF-FS:
# Description

The purpose of this is to create a POC for initializing a publicly accessible airflow deployment for Mike Ward - This will be set up with a secure username/password, as well as being IP Whitelisted. Once I have verified this, I will be creating the same for both test/prod, and initializing the whitelisting.


NathanEne (approved):
LGTM

PR #874, opened 2026-06-17
Remove Enabled Key, Bump Airflow Version

LiamF-FS:
# Description

- Bump Airflow Version, remove Enabled key
https://github.com/apache/airflow/blob/helm-chart/1.21.0/chart/values.yaml


NathanEne (approved):
LGTM

PR #877, opened 2026-06-18
Tweak Dag Processor Bug

LiamF-FS:
Wrong format entered for resource requests - expects dict, not string.


Kashike-Umemura (approved):
Thank you~

PR #879, opened 2026-06-18
RM Hardcode of Email Addr

LiamF-FS:
# Description

Given the scope of the changes, I would like to be notified if DAGs are failing as well.
Going to keep the fail on purpose DAG for Mike Ward to see the dashboard.


Kashike-Umemura (approved):
You shall be notified. LGTM

PR #880, opened 2026-06-22
Fix missing scheduler/triggerer expiration times

LiamF-FS:
- JWT Expiration Limits missing on some Test Airflow Microservices


Kashike-Umemura (approved):
Thank you for adding these :)

PR #883, opened 2026-07-02
Update Value Name, init replicas for PDF Generator

LiamF-FS:
# Description

- Up replicaCount for PDF Generator
- change `replica` -> `replicaCount`


AndersFarrFoundry (approved):
same same as other one!

PR #888, opened 2026-07-10
Fix PR Closed Action

LiamF-FS:
# Description

Migrations image was not being built as it is a dependency of the Airflow images. Therefore, we only need to clean up the backend/frontend/pdf-generator on merge in.


AndersFarrFoundry (approved):
Nice - this should fix that failing on merge action correct?

